PR #1, opened 2017-04-12
TBSTFNP-96 Update links and license year

henryju:


Extensions/SonarQube/license-terms.md
```diff
@@ -1,5 +1,5 @@
 The MIT License (MIT)
```
Evangelink: You will need to double check with @tomverin or Olivier but I think this can also be moved to LGPL.

PR #3, opened 2017-12-05
VSTS-117 Gulp build script

gregaubert:


tasks/scanner-cli/v3/task.json
```diff
@@ -0,0 +1,88 @@
+{
```
henryju: Looks like you have reformatted all json files. Is it intended? We are loosing all history on files...

gregaubert: Ouch, probably happened when I tested the update of the version on v3 tasks files to check if it worked. I'll check if there is an option to not break the current format. And revert those changes.

henryju: This is not a big deal if format is changed when packaging the extension

gregaubert: Currently it's updating the original file, maybe I misunderstood and we want to update only the packed file ?

henryju: In the original script, the task.json files were commited in the repo with version 0.0.0 and never updated. Only the packed file was updated before calling tfx

PR #4, opened 2017-12-06
Feature/jh/sonarcloud

henryju:


tasks/scanner-cli/v3/task.json
```diff
@@ -1,8 +1,8 @@
 {
     "id": "9F57024B-31F9-4E58-9E39-A47CCC098F03",
     "name": "SonarQubeScannerCli",
-    "friendlyName": "SonarQube  Scanner CLI",
-    "description": "Perform a SonarQube analysis of the source code of your PHP, JavaScript... projects",
+    "friendlyName": "Scanner CLI",
+    "description": "Perform a SonarQube or SonarCloud analysis of the source code of your PHP, JavaScript... projects",
     "helpMarkDown": "[More Information](http://redirect.sonarsource.com/doc/install-configure-scanner-tfs-ts.html)",
```
gregaubert: Maybe `Version: 3.0.2.` shoule be added here for every v3 tasks, it was added by the build script before.

vss-extension.json
```diff
@@ -134,7 +142,59 @@
...
+                "url": {
+                    "displayName": "Server Url",
+                    "value": "https://sonarcloud.io",
+                    "isVisible": false
+                },
+                "dataSources": [
+                    {
+                      "name": "TestConnection",
+                      "endpointUrl": "{{endpoint.url}}/api/organizations/search_my_organizations",
+                      "resultSelector": "jsonpath:$.organizations[*]"
```
gregaubert: I think you are using 4 spaces tabs normally. I don't know what editor you are using and how you format ts, js or json file but maybe we could use prettier (formatting can be done automatically on save in vscode), that's what we are using in SonarQube.

tasks/scanner-cli/v3/task.json
henryju: I don't like duplicating the version manually. In case we need to update tasks in the future, this is likely we will forgot. What about doing it automatically with gulp?

vss-extension.json
henryju: +1 to use prettier. Do we have a shared configuration somewhere?

gregaubert: Ok cool, the prettier configuration is in the `package.json`, on SonarQube we usually follow the prettier "trend" (nearly no custom rules). And if a rule change a lot from a version to another and we want it we just do a commit with the new formatting.
And for vscode, I use : 
```
  "editor.formatOnType": false,
  "editor.formatOnSave": true, 
  "javascript.validate.enable": false,
  "javascript.format.enable": false,
  "files.autoSave": "afterDelay",
  "files.autoSaveDelay": 10000,
```

tasks/scanner-cli/v3/task.json
gregaubert: Ok, so what do you think ?
For v3 taking the current version field of the task and use that to update `helpMarkDown` ?
Or add a new parameter like `--v3Version 3.0.15` for exemple ?

henryju: I would take current version of the task if possible.

PR #15, opened 2018-01-09
Fix migration from v3

henryju:


extensions/sonarqube/tasks/prepare/new/task.json
```diff
@@ -77,11 +77,11 @@
       "label": "Project Key",
       "required": true,
       "helpMarkDown": "The SonarQube project unique key, i.e. `sonar.projectKey`",
+      "visibleRule": "scannerMode = CLI && configMode = manual",
```
gregaubert: This line should be removed

PR #22, opened 2018-02-19
Fix bug in request logging and add debug logs

gregaubert:


henryju:
Maybe create a JIRA ticket so that we can do a release with these changes

gregaubert:
I created https://jira.sonarsource.com/browse/VSTS-138

PR #26, opened 2018-03-08
Embed Dotnet Scanner for MSBuild (4.1.0.1148)

michalb-sonar:
Fixes https://jira.sonarsource.com/browse/VSTS-145
Fixes https://jira.sonarsource.com/browse/VSTS-146

### DIFF
--- a/common/ts/sonarqube/Scanner.ts
+++ b/common/ts/sonarqube/Scanner.ts
@@ -81,13 +81,14 @@ export class ScannerCLI extends Scanner {
   }
 
   public async runAnalysis() {
-    let scannerExe = tl.resolve(this.rootPath, 'sonar-scanner', 'bin', 'sonar-scanner');
+    let scannerCliScripts = tl.resolve(this.rootPath, 'sonar-scanner', 'bin', 'sonar-scanner');
+    
     if (isWindows()) {
-      scannerExe += '.bat';
+      scannerCliScripts += '.bat';
     } else {
-      await fs.chmod(scannerExe, '777');
+      await fs.chmod(scannerCliScripts, '777');
     }
-    const scannerRunner = tl.tool(scannerExe);
+    const scannerRunner = tl.tool(scannerCliScripts);
     await scannerRunner.exec();
   }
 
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -39,32 +39,48 @@ gulp.task('npminstall', () =>
...
+    path.join(paths.build.extensions.sonarqubeTasks,  'prepare', 'new', 'dotnet-sonar-scanner-msbuild'),   
+    path.join(paths.build.extensions.sonarcloudTasks, 'prepare', 'new', 'dotnet-sonar-scanner-msbuild')    
   ];
+
   const cliFolders = [
-    path.join(paths.build.extensions.sonarqubeTasks, 'scanner-cli', 'old', 'sonar-scanner'),
-    path.join(paths.build.extensions.sonarqubeTasks, 'analyze', 'new', 'sonar-scanner'),
-    path.join(paths.build.extensions.sonarcloudTasks, 'analyze', 'new', 'sonar-scanner')
+    path.join(paths.build.extensions.sonarqubeTasks,  'scanner-cli', 'old', 'sonar-scanner'),
+    path.join(paths.build.extensions.sonarcloudTasks, 'analyze',     'new', 'sonar-scanner')
```
michalb-sonar: The sonarqube vsix contains scanner-cli in folder `dotnet-sonar-scanner`. Debugging to see where it comes from.
Otherwise, the msbuild/cli scanner files are where I'd expect them to be.

common/ts/sonarqube/Scanner.ts
```diff
@@ -125,41 +125,65 @@ export class ScannerMSBuild extends Scanner {
...
+    await fs.chmod(scannerCliShellScripts, '777');
+  }
+
+  private getScannerRunner(scannerPath : string, isExeScanner : boolean) {
+    if (isExeScanner) {
+      return tl.tool(scannerPath);
     }
+
+    const dotnetToolPath = tl.which('dotnet', true);
+    let scannerRunner = tl.tool(dotnetToolPath);
```
henryju: Can be const

gulpfile.js
```diff
@@ -39,32 +39,48 @@ gulp.task('npminstall', () =>
     .pipe(es.mapSync(file => npmInstallTask(file.path)))
 );
 
-gulp.task('scanner:download', () =>
-  download(scanner.url)
-    .pipe(decompress())
-    .pipe(gulp.dest(paths.build.scanner))
+gulp.task('scanner:download', () => {
```
henryju: My little understanding of gulp is that you should create 2 subtasks, and make this task depends on the two sub tasks. If you don't, since "download" is asynchronous, you may end up having only one completed.

henryju: Or use es.merge as below

michalb-sonar: Thanks, that was it! I'm surprised the packaging task didn't fail when files to copy were not present on the disk.

PR #28, opened 2018-03-14
MMF-1123 Quality Widgets

gregaubert:


gregaubert:
@henryju If you can just have a quick look just in case you can spot anything weird :)

gulpfile.js
```diff
@@ -191,7 +191,11 @@ gulp.task('extension:copy', () =>
   es
     .merge(
       gulp.src(
-        path.join(paths.extensions.root, '**', '@(vss-extension.json|extension-icon.png|*.md)')
+        path.join(
+          paths.extensions.root,
+          '**',
+          '@(vss-extension.json|extension-icon.png|*.md|*.html)'
```
henryju: I don't see any html file added to the extension, so why do you need to add that?

extensions/sonarcloud/vss-extension.json
```diff
@@ -53,6 +53,7 @@
     "type": "git",
     "uri": "https://github.com/SonarSource/sonar-scanner-vsts"
   },
+  "baseUri": "https://sonarcloud.io",
```
henryju: Have you checked it doesn't break other resources (like icons or tasks)?

gulpfile.js
gregaubert: Ah thanks it's a leftover from one of my early testing!

extensions/sonarcloud/vss-extension.json
gregaubert: I'll check again to be sure! Thanks!

gregaubert: Seems to be all good!

PR #29, opened 2018-03-31
Added build templates

duncanp-sonar:
Fixed [VSTS-128](https://jira.sonarsource.com/browse/VSTS-128)

### DIFF
--- a/extensions/sonarcloud/templates/gradle/template.json
+++ b/extensions/sonarcloud/templates/gradle/template.json
@@ -12,7 +12,6 @@
         "enabled": true,
         "continueOnError": false,
         "alwaysRun": false,
-        "displayName": "Prepare analysis on SonarCloud",
         "timeoutInMinutes": 0,
         "task": {
           "id": "14d9cde6-c1da-4d55-aa01-2965cd301255",
@@ -48,14 +47,13 @@
         "enabled": false,
         "continueOnError": false,
         "alwaysRun": false,
-        "displayName": "Publish Quality Gate Result",
         "timeoutInMinutes": 0,
         "task": {
           "id": "38b27399-a642-40af-bb7d-9971f69712e8",
           "versionSpec": "1.*",
           "definitionType": "task"
         },
-        "inputs": { "pollingTimeoutSec": "300" }
+        "inputs": {}
       }
     ],
     "options": [
--- a/extensions/sonarcloud/templates/gradle/template.json
+++ b/extensions/sonarcloud/templates/gradle/template.json
@@ -1,0 +1,109 @@
+{
+  "id": "sonarcloud_gradle_template",
+  "name": "Gradle with SonarCloud",
+  "category": "Build",
+  "iconTaskId": "ce096e50-6155-4de8-8800-4221aaeed4a1",
+  "defaultHostedQueue": null,
+  "description": "Build your Java projects and run tests with Gradle using a Gradle wrapper script, and analyze with SonarCloud.",
+  "template": {
+    "build": [
+      {
+        "environment": {},
+        "enabled": true,
+        "continueOnError": false,
+        "alwaysRun": false,
+        "displayName": "Prepare analysis on SonarCloud",
+        "timeoutInMinutes": 0,
+        "task": {
+          "id": "14d9cde6-c1da-4d55-aa01-2965cd301255",
+          "versionSpec": "1.*",
+          "definitionType": "task"
+        },
+        "inputs": {
+          "scannerMode": "Other",
+          "configMode": "file",
+          "configFile": "sonar-project.properties",
+          "cliProjectKey": "",
+          "projectKey": "",
+          "cliProjectName": "",
+          "projectName": "",
+          "cliProjectVersion": "1.0",
+          "projectVersion": "1.0",
+          "cliSources": ".",
+          "extraProperties": "# Additional properties that will be passed to the scanner, \n# Put one key=value per line, example:\n# sonar.exclusions=**/*.bin"
+        }
+      },
+      {
+        "environment": {},
+        "enabled": true,
+        "continueOnError": false,
+        "alwaysRun": false,
+        "timeoutInMinutes": 0,
+        "task": {
+          "id": "8d8eebd8-2b94-4c97-85af-839254cc6da4",
+          "versionSpec": "2.*",
+          "definitionType": null
+        },
+        "inputs": {
+          "wrapperScript": "$(Parameters.wrapperScript)",
+          "tasks": "$(Parameters.tasks)",
+          "codeCoverageTool": "JaCoCo",
+          "sqAnalysisEnabled": "true",
+          "sqGradlePluginVersionChoice": "specify",
+          "sqGradlePluginVersion": "2.6.1"
+        }
+      },
+      {
+        "environment": {},
+        "enabled": false,
+        "continueOnError": false,
+        "alwaysRun": false,
+        "displayName": "Publish Quality Gate Result",
+        "timeoutInMinutes": 0,
+        "task": {
+          "id": "38b27399-a642-40af-bb7d-9971f69712e8",
+          "versionSpec": "1.*",
+          "definitionType": "task"
+        },
+        "inputs": { "pollingTimeoutSec": "300" }
+      }
+    ],
+    "options": [
+      {
+        "enabled": true,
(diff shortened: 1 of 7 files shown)


extensions/sonarcloud/templates/gradle/template.json
```diff
@@ -0,0 +1,109 @@
+{
+  "id": "sonarcloud_gradle_template",
```
duncanp-sonar: A build template is just a _.json_ file listing the tasks to be executed, and the property values to be pre-set for those tasks. The template file must be called _template.json_. There is no executable code.
I've added four templates: Maven, Gradle, VS Build and .NET Core build.

extensions/sonarcloud/templates/gradle/template.json
```diff
@@ -0,0 +1,109 @@
...
+        "timeoutInMinutes": 0,
+        "task": {
+          "id": "14d9cde6-c1da-4d55-aa01-2965cd301255",
+          "versionSpec": "1.*",
+          "definitionType": "task"
+        },
+        "inputs": {
+          "scannerMode": "Other",
+          "configMode": "file",
+          "configFile": "sonar-project.properties",
```
henryju: configMode and configFile are not visible when scannerMode = 'Other'. Do you really have to define a value in the template?

extensions/sonarcloud/templates/gradle/template.json
```diff
@@ -0,0 +1,109 @@
...
+          "cliProjectKey": "",
+          "projectKey": "",
+          "cliProjectName": "",
+          "projectName": "",
+          "cliProjectVersion": "1.0",
+          "projectVersion": "1.0",
+          "cliSources": ".",
```
henryju: Same for all the previous props. I would find it more readable to not specify them in the template (if possible).

extensions/sonarcloud/templates/gradle/template.json
```diff
@@ -0,0 +1,109 @@
...
+          "versionSpec": "2.*",
+          "definitionType": null
+        },
+        "inputs": {
+          "wrapperScript": "$(Parameters.wrapperScript)",
+          "tasks": "$(Parameters.tasks)",
+          "codeCoverageTool": "JaCoCo",
+          "sqAnalysisEnabled": "true",
+          "sqGradlePluginVersionChoice": "specify",
+          "sqGradlePluginVersion": "2.6.1"
```
henryju: 2.6.2 is the latest version

extensions/sonarcloud/templates/maven/template.json
```diff
@@ -0,0 +1,104 @@
...
+          "scannerMode": "Other",
+          "configMode": "file",
+          "configFile": "sonar-project.properties",
+          "cliProjectKey": "",
+          "projectKey": "",
+          "cliProjectName": "",
+          "projectName": "",
+          "cliProjectVersion": "1.0",
+          "projectVersion": "1.0",
+          "cliSources": ".",
```
henryju: Same comment than for Gradle, don't specify hidden props if possible.

extensions/sonarcloud/templates/maven/template.json
```diff
@@ -0,0 +1,104 @@
...
+        "alwaysRun": false,
+        "timeoutInMinutes": 0,
+        "task": {
+          "id": "ac4ee482-65da-4485-a532-7b085873e532",
+          "versionSpec": "2.*",
+          "definitionType": null
+        },
+        "inputs": {
+          "mavenPOMFile": "$(Parameters.mavenPOMFile)",
+          "goals": "package",
```
henryju: I wonder if it is good to specify values here. If you don't, will it fallback to use default value? If yes, then in will be better to not fix the value in the template, but instead rely on the task defaults. This way, if they later do some changes to their default, our template will stay "up to date".

extensions/sonarcloud/templates/gradle/template.json
duncanp-sonar: Done

extensions/sonarcloud/templates/gradle/template.json
duncanp-sonar: I've removed all of the properties except "scannerMode" from all of the templates.

extensions/sonarcloud/templates/maven/template.json
duncanp-sonar: I've removed _goals_ but left _mavenPOMFile_ as it refers to a process level-variable.

extensions/sonarcloud/templates/gradle/template.json
```diff
@@ -0,0 +1,99 @@
...
+        "continueOnError": false,
+        "alwaysRun": false,
+        "displayName": "Publish Quality Gate Result",
+        "timeoutInMinutes": 0,
+        "task": {
+          "id": "38b27399-a642-40af-bb7d-9971f69712e8",
+          "versionSpec": "1.*",
+          "definitionType": "task"
+        },
+        "inputs": { "pollingTimeoutSec": "300" }
```
henryju: What about removing this input from the template?

extensions/sonarcloud/templates/maven/template.json
```diff
@@ -0,0 +1,91 @@
...
+        "continueOnError": false,
+        "alwaysRun": false,
+        "displayName": "Publish Quality Gate Result",
+        "timeoutInMinutes": 0,
+        "task": {
+          "id": "38b27399-a642-40af-bb7d-9971f69712e8",
+          "versionSpec": "1.*",
+          "definitionType": "task"
+        },
+        "inputs": { "pollingTimeoutSec": "300" }
```
henryju: Same

extensions/sonarcloud/templates/netcore/template.json
```diff
@@ -0,0 +1,163 @@
...
+        "displayName": "Publish Quality Gate Result",
+        "timeoutInMinutes": 0,
+        "condition": "succeeded()",
+        "task": {
+          "id": "38b27399-a642-40af-bb7d-9971f69712e8",
+          "versionSpec": "1.*",
+          "definitionType": "task"
+        },
+        "inputs": {
+          "pollingTimeoutSec": "300"
```
henryju: Same

extensions/sonarcloud/templates/netcore/template.json
```diff
@@ -0,0 +1,163 @@
...
+          "definitionType": "task"
+        },
+        "inputs": {}
+      },
+      {
+        "environment": {},
+        "enabled": false,
+        "continueOnError": false,
+        "alwaysRun": false,
+        "displayName": "Publish Quality Gate Result",
```
henryju: Same than for inputs, what if you don't specify a displayName?

extensions/sonarcloud/templates/netdesktop/template.json
```diff
@@ -0,0 +1,167 @@
...
+        "displayName": "Publish Quality Gate Result",
+        "timeoutInMinutes": 0,
+        "condition": "succeeded()",
+        "task": {
+          "id": "38b27399-a642-40af-bb7d-9971f69712e8",
+          "versionSpec": "1.*",
+          "definitionType": "task"
+        },
+        "inputs": {
+          "pollingTimeoutSec": "300"
```
henryju: same

extensions/sonarcloud/templates/gradle/template.json
duncanp-sonar: Done

extensions/sonarcloud/templates/netcore/template.json
duncanp-sonar: It seems to use the _friendlyName_ from the _task.json_. I've removed the _displayName_ for all of our tasks.
I've left the display names for the Microsoft tasks in cases where the corresponding default VSTS template has a display name (i.e. for the NuGet restore step and _dotnet_ tasks).

henryju: I would also remove for Microsoft tasks except if we see a benefit to not show the default friendly name. But that's minor.

duncanp-sonar: All of our templates are closely based on one of the default existing VSTS templates. I deliberately didn't change the definitions of the tasks copied from the standard templates to make it easier to compare our templates against the default templates in the future.

As it happens, the display name is required in three out of the four Microsoft tasks (the same task is run multiple times but with different parameters).

PR #30, opened 2018-04-08
Allow Multiple Scans To Be Published

vanderby:
This will allow multiple scans to be published to the build summary. A use case would be an Angular project with .NET backend where one scan would come from MSBUILD and the other from the sonar-scanner on the TypeScript front end.

I also added placing the project name next to the quality gate to differentiate which report belongs to which gate.

image

### DIFF
--- a/common/ts/publish-task.ts
+++ b/common/ts/publish-task.ts
@@ -22,28 +22,35 @@ export default async function publishTask(endpointType: EndpointType) {
   const endpoint = new Endpoint(endpointData.type, endpointData.data);
   const metrics = await Metrics.getAllMetrics(endpoint);
 
-  const taskReport = await TaskReport.createTaskReportFromFile();
   const timeoutSec = timeoutInSeconds();
-  try {
-    const task = await Task.waitForTaskCompletion(endpoint, taskReport.ceTaskId, timeoutSec);
-    const analysis = await Analysis.getAnalysis(
-      task.analysisId,
-      endpoint,
-      metrics,
-      taskReport.dashboardUrl
-    );
-    publishBuildSummary(analysis.getHtmlAnalysisReport(), endpoint.type);
-  } catch (e) {
-    if (e instanceof TimeOutReachedError) {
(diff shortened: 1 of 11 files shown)


henryju:
Sound interesting. @gregaubert can you have a look when you have time?

common/ts/publish-task.ts
```diff
@@ -22,28 +22,36 @@ export default async function publishTask(endpointType: EndpointType) {
...
+        } else {
+          throw e;
+        }
+      }
+    });
+  });
+
+  const buildSummaries = await Promise.all(buildSummariesPromises);
+  const reducer = (accumulator: string, currentValue: string) => `${accumulator}\r\n${currentValue}`;
+  publishBuildSummary(buildSummaries.reduce(reducer), endpoint.type);
```
gregaubert: And with the previous change, you don't need that reduce function and you can just use `Array.join` on the analyses:
```ts
publishBuildSummary(analyses.join('\r\n'), endpoint.type);
```

common/ts/sonarqube/Analysis.ts
```diff
@@ -62,7 +63,7 @@ export default class Analysis {
       font-size: 12px;
       margin-left: 15px;`;
     return `<div style="padding-top: 8px;">
-      <span>Quality Gate</span>
+      <span>${this.projectName} Quality Gate</span>
```
gregaubert: We don't want to add the project name if there is only one analysis to be displayed.

vanderby:
Updated per the PR comments.

gregaubert:
Thanks for the update! You still need to rebase on feature/greg/fix_build so we can have that green build and tests running.

gregaubert:
Indeed there is another failure due to the PR being external. I cherry picked your changes and created a new internal PR to run the analysis. I'll do a few more tests when I have time and then merge it.

gregaubert:
Changes have been merged from the other PR, so I'll close this one. Thanks for your contribution! :+1:

PR #38, opened 2018-04-24
Allow Multiple Scans To Be Published

gregaubert:
Copy of PR https://github.com/SonarSource/sonar-scanner-vsts/pull/30

### DIFF
--- a/common/ts/package.json
+++ b/common/ts/package.json
@@ -6,8 +6,8 @@
   "license": "LGPL-3.0",
   "dependencies": {
     "fs-extra": "5.0.0",
-    "request": "2.83.0",
-    "vsts-task-lib": "2.3.0",
+    "request": "2.85.0",
+    "vsts-task-lib": "2.4.0",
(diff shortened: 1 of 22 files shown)


common/ts/package.json
```diff
@@ -6,9 +6,9 @@
   "license": "LGPL-3.0",
   "dependencies": {
     "fs-extra": "5.0.0",
-    "request": "2.83.0",
-    "vsts-task-lib": "2.3.0",
-    "vso-node-api": "6.3.2"
+    "request": "2.85.0",
+    "vsts-task-lib": "2.4.0",
```
henryju: Do you master the implication of updating those libraries?

gregaubert: And I just noticed that they removed some apis from `vso-node-api` so I reverted the update for this one, just in case.
I think we can keep the updates of `request` and `vsts-task-lib`, they have not changed much, no breaking changes. And the app is running as before with their updates.

PR #39, opened 2018-05-28
VSTS-156 Enable branch/PR configuration on SonarQube

henryju:
WIP


common/ts/prepare-task.ts
```diff
@@ -12,7 +12,7 @@ export default async function prepareTask(endpoint: Endpoint, rootPath: string)
 
   const props: { [key: string]: string } = {};
 
-  if (endpoint.type === EndpointType.SonarCloud) {
+  if (endpoint.type === EndpointType.SonarCloud || branchFeatureSupported(endpoint)) {
```
gjambet: <nitpicking>for the sake of symmetry I would also extract the sonarcloud check in a small method</nitpicking>

common/ts/prepare-task.ts
```diff
@@ -37,6 +37,14 @@ export default async function prepareTask(endpoint: Endpoint, rootPath: string)
   await scanner.runPrepare();
 }
 
+function branchFeatureSupported(endpoint) {
+  if (endpoint.type === EndpointType.SonarCloud) {
+    return true;
+  }
+  // TODO check SQ version >= 7.2 and commercial branch feature available
```
Daniel-Svensson: Does this mean that sonarqube extension will start requiring the "commercial branch feature" for pull request comment integrations with once version 4 of the plugins supports it?

henryju: @Daniel-Svensson we are still discussing the options, but our current plan is:
* don't change the behavior when SQ < 7.2 = don't try to pass any parameter.
* when we detect SQ >= 7.2, pass automatically all sonar.branch.* or sonar.pullrequest.* parameters that are needed to have proper branch/PR decoration support. But you're right, this depends on the commercial feature.

Is it a problem for you?

common/ts/helpers/request.ts
```diff
@@ -26,18 +27,59 @@ export function getJSON(endpoint: Endpoint, path: string, query?: RequestData):
...
+            `[SQ] API GET '${path}' failed, status code was: ${response.statusCode}`
+          );
         }
         return resolve(body || {});
       }
     );
   });
 }
+
+export function getServerVersion(endpoint: Endpoint): Promise<semver.SemVer> {
```
gregaubert: I think you are basically duplicating all the code from `getJSON` (expect for the `json: true` option). So maybe it would be better to create a `get` method with nearly the same signature as `getJSON`, it could also take some optional options (so that `getJSON`, could use it and pass `{json: true}` as an option) to prevent code duplication.

And then you could use it like this: 
```ts
get(endpoint, '/api/server/version').then(semver.coerce)
```

common/ts/prepare-task.ts
Daniel-Svensson: Currently we use sonarqube for the pull request comment support (as described on https://docs.sonarqube.org/display/SCAN/SonarQube+Extension+3.0) using the free version.
This depends on "sonar.analysis.mode=preview" as I understand it, will that possibility be removed in 7.2 ? It is a bit sad too loose such a feature when upgrading the vsts extension from v3.

While we have discussed uppgrading and maybe bying developer edition that is a process which is not very fast (to get approval). Loosing pr commenting might have the effect of stopping such effort. 

Continue making a nice Product. Features such as 2-way integrationg (changing status etc) between VSTS comments and sonarqube server would definitly help motivate paying for the product

henryju:
@gregaubert Feedback applied

henryju:
@gregaubert Good catch. Fixed

PR #47, opened 2018-08-01
Added support spaces in build path

tasadar2:


henryju:
@gregaubert Can you have a look please

PR #55, opened 2018-11-14
Update request.ts for SELF_SIGNED_CERT_IN_CHAIN node error

peter-mourfield:
set rejectUnauthorized: false so that the Azure DevOps scanner can communicate with an instance of SonarQube that uses self-signed certificates. This avoids the node error SELF_SIGNED_CERT_IN_CHAIN

### DIFF
--- a/common/ts/helpers/request.ts
+++ b/common/ts/helpers/request.ts
@@ -16,6 +16,7 @@ function get(endpoint: Endpoint, path: string, isJson: boolean, query?: RequestD
     if (query) {
       options.qs = query;
       options.useQuerystring = true;
+      options.rejectUnauthorized = false;
     }
     request.get(
       {


peter-mourfield:
Looking back I believe this should be set via a sonar.* property that users could disable based on their security posture, defaulted to true obviously.

pierre-guillot-sonarsource:
Thanks for submitting this PR. As there is no traction on this for a long time, i'm closing it. If this is still a feature valuable for you, please consider posting in our [community forum](https://community.sonarsource.com/c/suggestions/features) a feature request.

NorDroN:
Hi!
I have the same problem.
Why this PR was rejected?
[Community forum](https://community.sonarsource.com/t/self-signed-cert-in-chain-error-on-prepare-analysis-step-when-using-azure-devops-and-internal-sonarqube-server/4257)

PR #60, opened 2019-02-21
Fix prepare task when using SonarCloud and MSBuild scanner

twaalewijn:
Fixes the issue I reported on the [community](https://community.sonarsource.com/t/azure-devops-sonarcloud-prepare-analysis-configuration-task-ignores-selected-organization-when-fetching-quality-profiles/6812/3) site.

If you are using the MSBuild scanner you need to specify the organization through the `/o:` or `/organization:` argument when using SonarCloud.
Because this argument isn't specified the scanner always ends up downloading the default quality profiles.

I've tested this for the SonarCloud extension on my own Azure DevOps organization and verified that custom quality profiles are now correctly downloaded and used.

I didn't test what happens when using the SonarQube extension but I don't this fix has impact on there since it doesn't have an input field for selecting an organization.


twaalewijn:
Don't know why the build is failing.
It doesn't seem to be related to the changes I committed though I think.

PR #64, opened 2019-04-04
Add contributing guide

henryju:


README.md
```diff
@@ -1,35 +1,57 @@
...
+========================================
 
 [![Build Status](https://sonarsource.visualstudio.com/_apis/public/build/definitions/399fb241-ecc7-4802-8697-dcdd01fbb832/10/badge)](https://sonarsource.visualstudio.com/DotNetTeam%20Project/_build/index?definitionId=10)
 
-### License
+Marketplace
+-----------
 
-Copyright 2017-2018 SonarSource.
+https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarqube
```
duncanp-sonar: What about the SonarCloud extension?

README.md
```diff
@@ -1,35 +1,57 @@
-# SonarQube TFS/VSTS Marketplace Extension
+SonarQube TFS/VSTS Marketplace Extension
```
duncanp-sonar: Should probably say `Azure DevOps` rather than `TFS/VSTS`.

README.md
```diff
@@ -1,35 +1,57 @@
...
+--------------------------
 
-## Marketplace
+For support questions ("How do I?", "I got this error, why?", ...), please head to the [SonarSource forum](https://community.sonarsource.com/c/help). There are chances that a question similar to yours has already been answered. 
 
-https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarqube
+Be aware that this forum is a community, so the standard pleasantries ("Hi", "Thanks", ...) are expected. And if you don't get an answer to your thread, you should sit on your hands for at least three days before bumping it. Operators are not standing by. :-)
+
+
+Contributing
```
duncanp-sonar: This section looks like a copy of the new `contributing.md` document. Do we need both?

henryju: Not sure why it was duplicated in the first place. @janos-ss ?

README.md
henryju: Added

README.md
henryju: I put `Azure Pipelines` which seems more accurate.

README.md
janos-ss: Indeed there's some duplication here. I don't see a better alternative. Let me explain.

- The `contributing.md` document, when present, is highlighted when creating a pull request. I think we want that highlighted, as at that point the README will not jump in the face.

- In the README, it's good to have this text, as a potential contributor may miss the other file. If you discover the `contributing.md` document when you create a pull request, it's kinda too late.

To be honest I'm quite irritated by such duplication, but I think it's better this way. If you see a better alternative, I'm open to chat about it.

duncanp-sonar: Ok, makes sense. It sounds like the `contributing.md` is essential as that's part of the PR process.

The only alternative I can think of for the ReadMe would be to keep the `Contributing` header and have text saying something like "Please read the contribution guidelines _here_."... with a link to `contribution.md`.

janos-ss: I considered that, and I went with duplicating the full text.

PR #66, opened 2019-04-16
change demand

DaveSkender:
Update demanded agent capability from 'java' to 'JAVA_HOME' to be compliant with newer standards.

- See [Community bug fix](https://community.sonarsource.com/t/azure-devops-hosted-build-java-demanded-capability-not-present/8878)

### DIFF
--- a/extensions/sonarcloud/tasks/analyze/new/task.json
+++ b/extensions/sonarcloud/tasks/analyze/new/task.json
@@ -14,7 +14,7 @@
     "Patch": 0
   },
   "minimumAgentVersion": "2.119.1",
-  "demands": ["JAVA_HOME"],
+  "demands": [],
   "inputs": [],
   "execution": {
     "Node": {
--- a/extensions/sonarcloud/tasks/analyze/new/task.json
+++ b/extensions/sonarcloud/tasks/analyze/new/task.json
@@ -14,7 +14,7 @@
     "Patch": 0
   },
   "minimumAgentVersion": "2.119.1",
-  "demands": ["java"],
+  "demands": ["JAVA_HOME"],
   "inputs": [],
   "execution": {
     "Node": {
--- a/extensions/sonarqube/tasks/analyze/new/task.json
+++ b/extensions/sonarqube/tasks/analyze/new/task.json
@@ -15,7 +15,7 @@
   },
   "releaseNotes": "To be used with the new version of the `Prepare Analysis Configuration` task.",
   "minimumAgentVersion": "2.119.1",
-  "demands": ["JAVA_HOME"],
+  "demands": [],
   "inputs": [],
   "execution": {
     "Node": {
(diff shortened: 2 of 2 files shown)


DaveSkender:
Just realized that 'java' usually exists on Linux servers as a capability while 'JAVA_HOME' typically exists on Windows servers.  Seems like a catch 22 where if you supply one or the other you'll be excluding on OS all the time.  If there's no "OR" option for demands, it might be better to simply remove the demand and have a soft requirement that JAVA exists.

duncanp-sonar:
@DaveSkender thanks for identifying the issue and finding workarounds.

Based on your PR, we created [this](https://jira.sonarsource.com/browse/VSTS-195) bug ticket. However, having discussed it internally we don't think removing the Java demand is the right solution as it will break the mechanism Azure DevOps uses to allocate jobs to agents which will cause customer scenarios that currently work correctly to start failing.

Scenario: two builds, agent A that has all of the required capabilities and agent B that has all of the capabilities except Java. Both agents are in the same agent pool.

Currently, Azure DevOps will correctly allocate jobs only to agent A because it is the only one with the required capabilities. If we remove the demand for `java` from our extensions then Azure DevOps will start allocating jobs to agent B as well. Those jobs will fail. From an end user point of view, their jobs will arbitrarily started failing, depending on which agent the job is allocated to.

As discussed in your community thread, we think the underlying issue is a bug in the MS agent capability detection code, and the workaround in the meantime is to manually add the `java` demand when using self-hosted build agents that have a non-Oracle version of Java installed.

DaveSkender:
@duncanp-sonar it would be nice if Java vendors would have been consistent on naming the Java environment variable between Windows and Linux, right?  `java` on one and `JAVA_HOME` on the other is problematic.

Regarding agent pools, I manage our agent pools in Azure DevOps and would generally not mix Linux and Windows in the same pools.  Here's how we would prefer to handle it in my organization:

1. Identify need for Java on all servers to support SonarQube
2. Install Java to all build servers
3. Let people use SonarQube on all servers without needing to workaround by adding fake capabilities

The problem with the current approach is that the default condition is that it only works on Linux servers.  The workaround for Windows is to "fake a capability", which is messy and hacky IMO and takes people a while to figure it out after time spent troubleshooting.  Not good UX.

Whether or not your remove the demand, we're still going to manage our build server software dependencies.  I think Java is pretty basic, so it's not a problem to ensure it's included on my build servers.

duncanp-sonar:
@DaveSkender the environment variable names are a red herring. If you look at capability detection code in the Microsoft build agent repo the `java` capability is set for both Windows and non-Windows machines. The problem is that the build agent on Windows [doesn't correctly set it](https://github.com/microsoft/azure-pipelines-agent/issues/2046) for self-hosted agents using non-Oracle Java installations.

PR #67, opened 2019-04-25
VSTS-169 Show project name in QG build status summary

duncanp-sonar:
Fixed #169

Miscellaneous:
* refactored the publish-task to make it easier to test
* add test coverage for publish-task

### DIFF
--- a/common/ts/__tests__/publish-task-test.ts
+++ b/common/ts/__tests__/publish-task-test.ts
@@ -26,7 +26,7 @@ const METRICS = new Metrics([]);
 
 it('should fail unless SONARQUBE_SCANNER_PARAMS are supplied', async () => {
   jest.spyOn(tl, 'getVariable').mockImplementation(() => null);
-  jest.spyOn(tl, 'setResult');
+  jest.spyOn(tl, 'setResult').mockImplementation(() => null);
 
   await publishTask.default(EndpointType.SonarCloud);
 
--- a/common/ts/__tests__/publish-task-test.ts
+++ b/common/ts/__tests__/publish-task-test.ts
@@ -1,3 +1,5 @@
+import * as tl from 'vsts-task-lib/task';
+import { InvalidApiResourceVersionError } from 'vso-node-api/VsoClient';
 import Analysis from '../sonarqube/Analysis';
 import Endpoint, { EndpointType } from '../sonarqube/Endpoint';
 import Metrics from '../sonarqube/Metrics';
@@ -5,8 +7,6 @@ import Task, { TimeOutReachedError } from '../sonarqube/Task';
 import TaskReport from '../sonarqube/TaskReport';
 import * as publishTask from '../publish-task';
 import * as serverUtils from '../helpers/vsts-server-utils';
-import * as tl from 'vsts-task-lib/task';
-import { InvalidApiResourceVersionError } from 'vso-node-api/VsoClient';
 
 beforeEach(() => {
   jest.restoreAllMocks();
--- a/common/ts/__tests__/publish-task-test.ts
+++ b/common/ts/__tests__/publish-task-test.ts
@@ -1,0 +1,119 @@
+import Analysis from '../sonarqube/Analysis';
+import Endpoint, { EndpointType } from '../sonarqube/Endpoint';
+import Metrics from '../sonarqube/Metrics';
+import Task, { TimeOutReachedError } from '../sonarqube/Task';
+import TaskReport from '../sonarqube/TaskReport';
+import * as publishTask from '../publish-task';
+import * as serverUtils from '../helpers/vsts-server-utils';
+import * as tl from 'vsts-task-lib/task';
+import { InvalidApiResourceVersionError } from 'vso-node-api/VsoClient';
+
+beforeEach(() => {
+  jest.restoreAllMocks();
+});
+
+const TASK_REPORT = new TaskReport({
+  ceTaskId: '111',
+  ceTaskUrl: 'http://taskurl1',
+  dashboardUrl: 'http://dashboardurl1',
+  projectKey: 'projectKey1',
+  serverUrl: 'http:/serverUrl1'
+});
+
+const SC_ENDPOINT = new Endpoint(EndpointType.SonarCloud, { url: 'https://endpoint.url' });
+const SQ_ENDPOINT = new Endpoint(EndpointType.SonarQube, { url: 'https://endpoint.url' });
+const METRICS = new Metrics([]);
+
+it('should fail unless SONARQUBE_SCANNER_PARAMS are supplied', async () => {
+  jest.spyOn(tl, 'getVariable').mockImplementation(() => null);
+  jest.spyOn(tl, 'setResult')
+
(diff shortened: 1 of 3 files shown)


common/ts/publish-task.ts
```diff
@@ -25,29 +25,7 @@ export default async function publishTask(endpointType: EndpointType) {
   const timeoutSec = timeoutInSeconds();
   const taskReports = await TaskReport.createTaskReportsFromFiles();
   const analyses = await Promise.all(
-    taskReports.map(async taskReport => {
```
duncanp-sonar: I moved the code into a separate function to make it easier to test.

common/ts/publish-task.ts
```diff
@@ -56,3 +34,27 @@ export default async function publishTask(endpointType: EndpointType) {
...
+
+export async function getReportForTask(taskReport: TaskReport, metrics: Metrics, endpoint: Endpoint, timeoutSec: number): Promise<string> {
+  try {
+    const task = await Task.waitForTaskCompletion(endpoint, taskReport.ceTaskId, timeoutSec);
+    const analysis = await Analysis.getAnalysis({
+      analysisId: task.analysisId,
+      dashboardUrl: taskReport.dashboardUrl,
+      endpoint,
+      metrics,
+      projectName: task.componentName
```
duncanp-sonar: This is the only line of the product code that has changed. Previously there was a conditional check so the project name was only included if there were multiple projects.

common/ts/sonarqube/__tests__/TaskReport-test.ts
```diff
@@ -23,3 +28,57 @@ serverUrl=http://sonar`,
 
   tmpReport.removeCallback();
 });
+
+
+it('should parse all reports', async () => {
```
duncanp-sonar: Miscellaneous: these tests check the pre-existing behaviour i.e. not affected the fix. I added these tests mainly to build familiarity with the test and mocking frameworks.

common/ts/__tests__/publish-task-test.ts
```diff
@@ -0,0 +1,119 @@
+import Analysis from '../sonarqube/Analysis';
```
duncanp-sonar: Most of these tests are for the pre-existing behaviour. I'll call out the check that covers the change made in the fix.

common/ts/__tests__/publish-task-test.ts
```diff
@@ -0,0 +1,119 @@
...
+  // Mock converting the Task into an html report
+  const returnedAnalysis = new Analysis({ status: '', conditions: [] }, EndpointType.SonarCloud, '', null, null);
+  jest.spyOn(Analysis, 'getAnalysis').mockImplementation(() => returnedAnalysis);
+  jest.spyOn(returnedAnalysis, 'getHtmlAnalysisReport').mockImplementation(() => 'dummy html');
+
+  const result = await publishTask.getReportForTask(TASK_REPORT, METRICS, SQ_ENDPOINT, 999);
+
+  expect(Task.waitForTaskCompletion).toHaveBeenCalledWith(SQ_ENDPOINT, TASK_REPORT.ceTaskId, 999);
+  expect(Analysis.getAnalysis).toBeCalledWith(
+    { analysisId: '123', dashboardUrl: 'http://dashboardurl1', endpoint: SQ_ENDPOINT, metrics: { metrics: [] }, projectName: 'componentName' });
```
duncanp-sonar: This test checks the new behaviour i.e. that the component/project name is always supplied when creating the Analysis object.

common/ts/publish-task.ts
henryju: In the end I'm not sure you should change that... it will make the widget take more space for "nothing". The PR https://github.com/SonarSource/sonar-scanner-vsts/pull/38 was already merged to fix VSTS-169. Not sure why the link was not made...

common/ts/__tests__/publish-task-test.ts
```diff
@@ -0,0 +1,119 @@
...
+  // Mock finding two report files to process
+  jest.spyOn(TaskReport, 'createTaskReportsFromFiles').mockImplementation(() => [TASK_REPORT, taskReport2]);
+
+  jest.spyOn(Task, 'waitForTaskCompletion').mockImplementation(() => { throw new TimeOutReachedError(); });
+  jest.spyOn(serverUtils, 'publishBuildSummary').mockImplementation(() => null);
+
+  await publishTask.default(EndpointType.SonarCloud);
+
+  expect(serverUtils.publishBuildSummary).toHaveBeenCalledTimes(1);
+  const publishSummaryMock = (serverUtils.publishBuildSummary as jest.Mock<any>);
```
gregaubert: I think you can also store what's returned by the `spyOn` (that you call on line 106) to have the spy/mock instance of that method.

common/ts/publish-task.ts
duncanp-sonar: @gregaubert did point that PR out to me. It's solving a slightly different problem: if multiple reports are found, then PR #38 displays them all in a single report which looks like this:

image

The build summary above was produced by a build that analyses an MSBuild project, two Java projects, and has a single _Publish_ step.

Unfortunately, there's a problem if multiple MSBuild projects are analysed, which is what VSTS-169 addresses. The report file for MSBuild projects is always written to the same place on disk, so the only way to see a summary for each project is to have multiple _publish_ steps, which is presumably what the user who reported the issue is doing. 
With this change, the summary reports look like this:

image

Without the project names, the summaries look identical.
Adding the project name doesn't change the vertical height of the widget. It might make it slightly wider, but there is nothing to the right of the widget that wouldn't doesn't matter.

This change is a simple fix to make the widgets look different. I've created a separate ticket to address the inconsistency in the handling of the reports: https://jira.sonarsource.com/browse/VSTS-191

common/ts/sonarqube/__tests__/TaskReport-test.ts
```diff
@@ -79,5 +79,10 @@ it('should find report files', async () => {
   expect(tl.getVariable).toBeCalledWith('Agent.BuildDirectory');
 
   expect(tl.findMatch).toHaveBeenCalledTimes(1);
-  expect(tl.findMatch).toBeCalledWith('mock root search path', '**\\report-task.txt');
+  expect(findMatchMock.mock.calls[0][0]).toBe('mock root search path');
+
+  // Match using a regular expression to take account of the different
+  // path separators used by "join" on Windows/non-Windows
+  const regEx = new RegExp(/^\*\*[\\/]report-task.txt$/);
```
gregaubert: I'm not sure it's better but you could also use [path.join](https://nodejs.org/api/path.html#path_path_join_paths)

duncanp-sonar: Doh! That's definitely simpler - not sure why I started messing about with regular expressions...

common/ts/publish-task.ts
henryju: Make totally sense, thanks for the explanation.

duncanp-sonar: Apologies, I meant to call this out in a PR comment but forgot :-(

PR #68, opened 2019-04-26
Fix scripts to also work on Windows machines

duncanp-sonar:


package.json
```diff
@@ -47,13 +47,13 @@
     "deploy": "./node_modules/.bin/gulp deploy",
     "sonarqube": "./node_modules/.bin/gulp sonarqube",
     "format":
-      "./node_modules/.bin/prettier --write --list-different '{common,extensions}/**/*.{js,ts}'",
+      "./node_modules/.bin/prettier --write --list-different \"{common,extensions}/**/*.{js,ts}\"",
```
duncanp-sonar: Changing single to double-quotes. I've tested the changes on Windows and on MacOS

PR #80, opened 2019-07-29
QA Hardening

mickael-caro-sonarsource:
Contains : 

VSTS-192 - Added code coverage on Next

VSTS-201 - An error message inside a unit test was displaying as error in Azure DevOps (expected, but ugly and worrysome in the UI). So a mock has been added, message has disappear.


package.json
```diff
@@ -33,8 +33,8 @@
     "prettier": "1.12.1",
     "request": "2.85.0",
     "semver": "5.5.0",
-    "sonarqube-scanner": "2.1.2",
-    "tfx-cli": "^0.7.6",
+    "sonarqube-scanner": "^2.5.0",
+    "tfx-cli": "^0.7.8",
```
gregaubert: In sonarcloud-core repo we like to have fixed dependencies (without the `^` before the version) to be sure to never have bad surprises. I don't know if it make sens for this repo but I just wanted to raise the point :)

package.json
```diff
@@ -53,25 +53,23 @@
...
+    "validate-ci": "npm run lint-report && npm run ts-check && npm run format-check && npm run test-ci"
   },
   "engines": {
     "node": ">=6"
   },
   "jest": {
     "clearMocks": true,
     "collectCoverageFrom": [
-      "**/*.ts"
+      "{common}**/*.ts"
```
gregaubert: You might want to update your `sonar.coverage.exclusions` setting for the sonar-scanner in `config/utils.js` to ignore things that are in the folder `extensions`

package.json
mickael-caro-sonarsource: Ok, done !

package.json
mickael-caro-sonarsource: And done as well, thanks !

PR #81, opened 2019-08-05
VSTS-190 Changed the directory for generating report-task.txt files

mickael-caro-sonarsource:
…and thus the directory to get report from for uploading summaries) + updated UT

### DIFF
--- a/common/ts/prepare-task.ts
+++ b/common/ts/prepare-task.ts
@@ -1,7 +1,7 @@
+import * as path from 'path';
 import * as semver from 'semver';
 import * as tl from 'azure-pipelines-task-lib/task';
 import * as vm from 'azure-devops-node-api';
-import * as path from 'path';
 import { Guid } from 'guid-typescript';
 import Endpoint, { EndpointType } from './sonarqube/Endpoint';
 import Scanner, { ScannerMode } from './sonarqube/Scanner';
--- a/common/ts/prepare-task.ts
+++ b/common/ts/prepare-task.ts
@@ -1,6 +1,7 @@
 import * as semver from 'semver';
 import * as tl from 'azure-pipelines-task-lib/task';
 import * as vm from 'azure-devops-node-api';
+import { Guid } from 'guid-typescript';
 import Endpoint, { EndpointType } from './sonarqube/Endpoint';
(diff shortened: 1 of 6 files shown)


common/ts/prepare-task.ts
```diff
@@ -118,6 +122,12 @@ function branchName(fullName: string) {
   return fullName;
 }
 
+export function reportPath(): string {
+  return `${tl.getVariable('Agent.TempDirectory')}\\${tl.getVariable(
```
duncanp-sonar: Is there a function that will build paths? (i.e. one that would handle using the correct path separators on different platforms?)
If not, it might be better to use "/" instead of  "\" as that is more Linux-friendly and Windows will handle both.

duncanp-sonar: I'm pretty sure the temp directory is deleted by the agent after each build so creating a sub-directory using the build ID is probably unnecessary. However, I can't find any documentation to confirm this and it's better to be safe.

It might be worth creating a `sonar` subdirectory to help with documentation/diagnosis e.g. if a user is trying to diagnose a build failure on their agent they probably won't know what is create those subfolders or why. However, if they saw the files were under a directory called `sonar` they'd be able to work it out.

i.e. make the root temp directory something like this:  {Agent.TempDirectory}\sonar\{BuildId}\

mickael-caro-sonarsource: Oh yes, totally forgot to use it here. Fixing it right now.

mickael-caro-sonarsource: The reason of using guid is just to handle multiple report-task.txt generated during a build (it's already supported for maven/gradle builds), so as long as we generate the path in the prepare task, and we don't even know the project key nor some kind of unique identifier of the currently built project, that's the pretty much only solution to achieve that. But i agree for the sonar directory, i can add it to the generated path.

duncanp-sonar: Absolutely, we need the `guid` subdir; I didn't mean to suggest we didn't. Sorry for not being clearer.

mickael-caro-sonarsource: Oh, ok, no problem, sorry too.

PR #82, opened 2019-08-07
VSTS-189 - Check QG Status in release task. 

mickael-caro-sonarsource:
Contains also  :

VSTS-208 
VSTS-209


common/ts/helpers/azdo-api-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+      });
+    });
+
+    tl.debug(JSON.stringify(patchBody));
+
+    const options = {
+      url:
+        collectionUri +
+        teamProjectId +
+        `_apis/build/builds/${buildId}/properties?api-version=5.0-preview.1`,
```
duncanp-sonar: Is there a non-preview API that we could use?

common/ts/publish-task.ts
```diff
@@ -28,13 +31,30 @@ export default async function publishTask(endpointType: EndpointType) {
     taskReports.map(taskReport => getReportForTask(taskReport, metrics, endpoint, timeoutSec))
   );
 
+  if (globalQualityGateStatus === '') {
```
duncanp-sonar: Does this mean we're treating _unknown_/_no analysis_ the same as _succeeded_?

extensions/sonarcloud/tasks/qgstatus/new/task.json
```diff
@@ -0,0 +1,34 @@
...
+  "preview": "true",
+  "inputs": [
+  ],
+  "instanceNameFormat": "Check SonarCloud Quality Gate Status",
+  "execution": {
+    "HttpRequest": {
+      "Execute": [
+        {
+          "RequestInputs": {
+            "EndpointUrl": "$(system.teamFoundationCollectionUri)$(system.teamProjectId)/_apis/build/builds/$(build.buildid)/properties?api-version=5.0-preview.1",
```
duncanp-sonar: (same question about non-preview API as above)

sonar-docs/analysis/scan/sonarscanner-for-azure-devops.md
```diff
@@ -52,7 +52,7 @@ Each extension provides three tasks you will use in your build definitions to an
 * **Run Code Analysis** task, to actually execute the analysis of the source code. 
    * This task is not required for Maven or Gradle projects, because scanner will be run as part of the Maven/Gradle build.
 * **Publish Quality Gate Result** task, to display the Quality Gate status in the build summary and give you a sense of whether the application is ready for production "quality-wise". 
-   * This tasks is optional. 
+   * This tasks is mandatory if you are using a release pipeline, with the {instance} quality gate status check pre-deployment gate, otherwise it's optional.
```
duncanp-sonar: Pre-existing typo: `This task is mandatory...` (_task_ should be singular)

duncanp-sonar: Is `{instance}` our custom markdown that will be resolved to SonarQube / SonarCloud when the docs are generated?
If so, is this documentation relevant for SonarQube at the moment since the new gate only seems to be for SonarCloud?

sonar-docs/analysis/scan/sonarscanner-for-azure-devops.md
```diff
@@ -148,17 +148,30 @@ When a build is run on a branch of your project, the extension automatically con
...
+<!-- sonarcloud -->
+## Using Release Pipelines
+You have the possibility to check the SonarCloud quality gate status in your release pipeline. It takes place as a [pre-deployment gate](https://docs.microsoft.com/en-us/azure/devops/pipelines/release/approvals/gates?view=azure-devops).
+
+1. In the **release pipeline**, add a stage, then click on **pre-deployment conditions**.
+2. Enable the **gates**, then click on add. Choose **SonarCloud QualityGate status check**
+3. Save your pipeline.
+
+**This feature is currently in preview, and the following notes are important** :
+* The **publish quality gate result** task has to be enabled in order to get this gate working.
```
duncanp-sonar: Should probably say that the `Publish Quality Gate Result` is in the build pipeline used to create the artefacts, just to avoid any misunderstanding.

Also, should probably use the correct capitialisation for the task name rather than being all lower-case.

sonar-docs/analysis/scan/sonarscanner-for-azure-devops.md
```diff
@@ -148,17 +148,30 @@ When a build is run on a branch of your project, the extension automatically con
...
+* If the quality gate is in the failed state, it will not be possible to get the pre-deployment gate passing as this status will remain in its initial state. You will have to execute another build with either the current issues corrected in SonarCloud, or with another commit for fixing them.
+* Please note also that current behavior of the pre-deployment gates in Release Pipelines check every 5 minutes the status, for a duration of 1 day by default. Knowing the fact that if the SonarCloud quality gate is failed and it will remains like this on Azure DevOps, you can down this duraction to 6min at least (so the gate will be evaluated only twice), or just cancel the release itself.
```
duncanp-sonar: "you can decrease this duration to a minimum of six minutes"

common/ts/publish-task.ts
mickael-caro-sonarsource: Yes, but you're right, i will add warn and unknown in this logic. Since the gate is yes / no, there will be no other kind of possible solution for the moment.

common/ts/helpers/azdo-api-utils.ts
mickael-caro-sonarsource: No, they are documented like that, even the TFS2018 U2 is in preview as well https://docs.microsoft.com/en-us/rest/api/azure/devops/build/properties/get%20build%20properties?view=azure-devops-rest-5.0

sonar-docs/analysis/scan/sonarscanner-for-azure-devops.md
mickael-caro-sonarsource: At first sight, yes, only for SonarCloud, but since the code has been simplified, i have to test with an Azure DevOps server and a SonarQube to see if it's ok, but to my mind it should be, so the first preview might be available for SQ as well.

mickael-caro-sonarsource: Thanks, this has been modified.

sonar-docs/analysis/scan/sonarscanner-for-azure-devops.md
mickael-caro-sonarsource: Yes of course, this has been modified.

sonar-docs/analysis/scan/sonarscanner-for-azure-devops.md
mickael-caro-sonarsource: This has been modified.

common/ts/publish-task.ts
```diff
@@ -28,13 +31,30 @@ export default async function publishTask(endpointType: EndpointType) {
...
+    globalQualityGateStatus = 'ok';
```
duncanp-sonar: There could be multiple analyses, and multiple task reports couldn't there? Won't the last task to be reported "win" in that case?

mickael-caro-sonarsource: The algorithm works like this : If one of the quality gate from the multiple analyses has an error / warn / none status then the global quality gate will fail, and that will not change if an 'ok' one appear after. To have the global status green, all of the analyses should have the 'OK' status, that's the reason of this code (variable not modified, so it's ok), see line 74/75/76 for the tests.

common/ts/publish-task.ts
```diff
@@ -28,13 +31,30 @@ export default async function publishTask(endpointType: EndpointType) {
...
+  }
+
+  fillBuildProperty();
```
duncanp-sonar: It would be worth adding some user-visible logging so users can see what is happening and to help us diagnose issues e.g. `Setting quality gate status for the build to xxx`

common/ts/__tests__/publish-task-test.ts
```diff
@@ -6,7 +6,7 @@ import Metrics from '../sonarqube/Metrics';
 import Task, { TimeOutReachedError } from '../sonarqube/Task';
 import TaskReport from '../sonarqube/TaskReport';
 import * as publishTask from '../publish-task';
-import * as serverUtils from '../helpers/vsts-server-utils';
+import * as serverUtils from '../helpers/azdo-server-utils';
```
duncanp-sonar: Are you planning on adding tests?

mickael-caro-sonarsource: Yes, it's in the pipe.

common/ts/publish-task.ts
```diff
@@ -35,7 +35,13 @@ export default async function publishTask(endpointType: EndpointType) {
     globalQualityGateStatus = 'ok';
   }
 
-  fillBuildProperty();
+  tl.debug(
+    `Following QualityGate status : '${globalQualityGateStatus}' has been deducted from ${
```
duncanp-sonar: It might be simpler to log two messages:
`Number of analyses in this build: {x}`
`Overall Quality Gate status: {x}`
(or perhaps "Calculated Overall Quality Gate status" or just "Calculated Quality Gate status" instead)
 
FYI it would be "deduced" rather than "deducted", although it would be more usual to say "calculated".

duncanp-sonar:
Latest changes look good to me.

common/ts/publish-task.ts
mickael-caro-sonarsource: Agreed, this has been implemented.

PR #83, opened 2019-08-13
VSTS-211 - Fixed a bug where if SONAR_SCANNER_OPTS is null after the …

mickael-caro-sonarsource:
…retrieval of its initial value, will fail the concat call.

### DIFF
--- a/common/ts/__tests__/prepare-task-test.ts
+++ b/common/ts/__tests__/prepare-task-test.ts
@@ -1,6 +1,7 @@
 import * as path from 'path';
 import * as tl from 'azure-pipelines-task-lib/task';
 import { Guid } from 'guid-typescript';
+import { when } from 'jest-when';
 import Endpoint, { EndpointType } from '../sonarqube/Endpoint';
 import * as prept from '../prepare-task';
 import * as request from '../helpers/request';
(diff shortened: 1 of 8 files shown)


common/ts/prepare-task.ts
```diff
@@ -54,6 +54,9 @@ export default async function prepareTask(endpoint: Endpoint, rootPath: string)
 
   if (scannerMode === ScannerMode.CLI) {
     let actualSonarOpts = tl.getVariable('SONAR_SCANNER_OPTS');
+    if (actualSonarOpts == null) {
```
gregaubert: No need to change it, but you can write short things like `let actualSonarOpts = tl.getVariable('SONAR_SCANNER_OPTS') || '';` for the same result

package.json
```diff
@@ -91,6 +91,7 @@
     "singleQuote": true
   },
   "dependencies": {
+    "@types/jest-when": "2.4.1",
```
gregaubert: You probably want this in `devDependencies` instead

common/ts/__tests__/publish-task-test.ts
```diff
@@ -60,8 +60,18 @@ it('check multiple report status and set global quality gate for build propertie
   taskReportArray.push(taskReport);
   taskReportArray.push(taskReport);
 
-  jest.spyOn(Task, 'waitForTaskCompletion').mockImplementation(() => returnedTaskOk);
-  jest.spyOn(TaskReport, 'createTaskReportsFromFiles').mockImplementation(() => taskReportArray);
+  jest.spyOn(Task, 'waitForTaskCompletion').mockImplementation(
+    () =>
+      new Promise<Task>(resolve => {
+        return resolve(returnedTaskOk);
```
gregaubert: You can do things like `jest.spyOn(Task, 'waitForTaskCompletion').mockResolvedValue(returnedTaskOk);` and it should have the same effect.

common/ts/__tests__/publish-task-test.ts
```diff
@@ -72,14 +82,29 @@ it('check multiple report status and set global quality gate for build propertie
     null
   );
 
-  jest.spyOn(Analysis, 'getAnalysis').mockImplementationOnce(() => returnedAnalysisOk);
-  jest.spyOn(Analysis, 'getAnalysis').mockImplementationOnce(() => returnedAnalysisOk);
+  jest.spyOn(Analysis, 'getAnalysis').mockImplementationOnce(
```
gregaubert: There is also a `jest.spyOn(Analysis, 'getAnalysis').mockResolvedValueOnce(returnedAnalysisOk);`

common/ts/__tests__/prepare-task-test.ts
```diff
@@ -24,8 +26,18 @@ it('should display warning for dedicated extension for Sonarcloud', async () =>
   jest.spyOn(tl, 'warning').mockImplementation(() => null);
   jest.spyOn(Scanner, 'getPrepareScanner').mockImplementation(() => scannerObject);
   jest.spyOn(scannerObject, 'runPrepare').mockImplementation(() => null);
-  jest.spyOn(request, 'getServerVersion').mockImplementation(() => '7.0.0');
-  jest.spyOn(prept, 'getDefaultBranch').mockImplementation(() => 'refs/heads/master');
+  jest.spyOn(request, 'getServerVersion').mockImplementation(
```
gregaubert: Same in this file.

common/ts/prepare-task.ts
O-heu: We get an error message with undefined, does this null check fixes that?

mickael-caro-sonarsource: I knew the syntax in C# but i was looking for something like this, thanks !

package.json
mickael-caro-sonarsource: Ok, i moved all dependencies in dev for jest stuff.

common/ts/__tests__/publish-task-test.ts
mickael-caro-sonarsource: Thanks, that works !

common/ts/prepare-task.ts
mickael-caro-sonarsource: Yes !

PR #85, opened 2019-08-13
VSTS-211: Cope with SONAR_SCANNER_OPTS not being defined.

samattridge:
When using the latest version of the Azure DevOps extension, if we don't specify a variable of `SONAR_SCANNER_OPTS` we see the following error during the `SonarQubePrepare` phase:

> ##[error]Cannot read property 'concat' of undefined

This change checks for the variable not be defined.


KevCJones:
Here to basically say +1
`##[error]Cannot read property 'concat' of undefined`
😭

KevCJones:
is the workaround here to define `SONAR_SCANNER_OPTS` then? Do i need to put anything in it?  concat make me think an empty array `[]`?

mickael-caro-sonarsource:
Hi, versions have been fixed right now, so i'm closing this pull request.

Thanks.

Mickaël

PR #88, opened 2019-09-06
VSTS-213 - Report-task.txt new location improvement

mickael-caro-sonarsource:
newest location of report-task.txt is now dedidated to either SonarCloud or SQ >= 7.2 + Added fallback research to previous folders in case of SQ < 7.2.

### DIFF
--- a/common/ts/prepare-task.ts
+++ b/common/ts/prepare-task.ts
@@ -30,6 +30,7 @@ export default async function prepareTask(endpoint: Endpoint, rootPath: string) 
 
   if (await branchFeatureSupported(endpoint)) {
     await populateBranchAndPrProps(props);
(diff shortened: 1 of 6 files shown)


common/ts/prepare-task.ts
```diff
@@ -30,6 +30,7 @@ export default async function prepareTask(endpoint: Endpoint, rootPath: string)
 
   if (await branchFeatureSupported(endpoint)) {
     await populateBranchAndPrProps(props);
+    props['sonar.scanner.metadataFilePath'] = reportPath();
```
henryju: I find it misleading. At least put a comment saying that branchFeatureSupported is by chance checking that version is > 7.2.

mickael-caro-sonarsource: Ok that has been added. Hardening soon to fix this.

PR #89, opened 2019-09-06
VSTS-172 - workaround with embedded find method

mickael-caro-sonarsource:
As discussed in the JIRA ticket, and as long as we are waiting for the fix on the lib side, both find and findMatch methods have been embedded to workaround the issue with the GC size limit exceeded.

### DIFF
--- a/common/ts/helpers/temp-find-method.ts
+++ b/common/ts/helpers/temp-find-method.ts
@@ -1,0 +1,380 @@
+/* istanbul ignore file */
+import * as tl from 'azure-pipelines-task-lib/task';
+import path = require('path');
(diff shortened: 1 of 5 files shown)


common/ts/helpers/temp-find-method.ts
```diff
@@ -0,0 +1,380 @@
+/* istanbul ignore file */
```
henryju: I would add a comment with a link to the GitHub issue, so that we know when to delete this file.

mickael-caro-sonarsource: Ok i added a message for that.

PR #91, opened 2019-09-10
VSTS-217 - Use Azure DevOps node api lib instead of request

mickael-caro-sonarsource:
For newly created AddBuildProperty method, request has been replace by the already referenced wrapper over AzDO REST API, which is easier to maintain.

### DIFF
--- a/common/ts/helpers/azdo-api-utils.ts
+++ b/common/ts/helpers/azdo-api-utils.ts
@@ -1,57 +1,54 @@
 import * as tl from 'azure-pipelines-task-lib/task';
-import * as request from 'request';
-import { getAuthToken } from './azdo-server-utils';
+import * as vm from 'azure-devops-node-api';
+import {
+  JsonPatchDocument,
+  JsonPatchOperation,
+  Operation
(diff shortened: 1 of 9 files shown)


common/ts/helpers/azdo-api-utils.ts
```diff
@@ -1,57 +1,54 @@
...
+      path: `/${property.propertyName}`,
+      value: `${property.propertyValue}`
     });
   });
+
+  tl.debug(JSON.stringify(patchBody));
+
+  const customHeader = { Authorization: `Bearer ${getAuthToken()}` };
+
+  const azdoWebApi = getWebApi(collectionUri);
```
gregaubert: You don't need to pass the team id anymore ?

mickael-caro-sonarsource: It's passed on line 38 now, directly in the updateBuildProperties method

gregaubert: Ah ok it's the `updateBuildProperties` that does the actual call...

PR #95, opened 2019-10-29
Update to Gulp 4

wouter-admiraal-sonarsource:
We use Node LTS in our Travis CI env. However, Gulp 3 is not compatible with NodeJS >=12.

We could downgrade our Travis CI env to Node 10, but I suggest we update Gulp, as we'll have to at some point anyway.

I didn't check the `deploy` tasks (obviously). `test`, ` build`, and `sonarqube` should work.

Please note that the tasks needed re-organizing, because Gulp now "parses" tasks top-down, and immediately throws an error if it encounters a task that it doesn't know about. This means you can no longer do stuff like:

```js
gulp.task('default', 'build');
gulp.task('build', () => { /* do stuff */ });
```

`build` being defined _after_ `default`, Gulp v4 will throw an error here. Instead, we need to make sure `build` is already defined.

Also note that the new recommended way is to create functions and export them. But I personally prefer the "old-ish" v3 syntax, which is still supported.


package.json
```diff
@@ -20,19 +21,19 @@
     "event-stream": "3.3.4",
     "fs-extra": "7.0.1",
     "globby": "7.1.1",
-    "gulp": "3.9.1",
+    "gulp": "^4.0.2",
```
gregaubert: Use a fixed version like the rest of the dependencies ?

gulpfile.js
```diff
@@ -139,7 +105,7 @@ gulp.task('tasks:new:ts', ['npminstall'], () =>
     .pipe(gulp.dest(paths.build.extensions.root))
 );
 
-gulp.task('tasks:new:common:ts', ['npminstall'], () => {
+gulp.task('tasks:new:common:ts', gulp.parallel('npminstall'), () => {
```
gregaubert: What's the effect of `gulp.parallel` on a single task ?

wouter-admiraal-sonarsource:
> Seems good. But that's really hard to review, there is a mix of changes and everything that moved inside the file, so there is now way to know what "just moved" and what "moved and changed"...

Yeah, sorry about that. Do you want me to split it into several commits? One with the API changes, and one with the re-ordering? I can try and rollback 🙂 .

gulpfile.js
wouter-admiraal-sonarsource: Nothing, I guess. It will just trigger a single task anyway. Could be series as well. It's the result of a search/replace I did in the code 🤷‍♂. I didn't find what the best practice is when depending on a single task.

gregaubert: Ahh ok, so that's because you can not pass just a single task name anymore ?

gregaubert:
> Yeah, sorry about that. Do you want me to split it into several commits? One with the API changes, and one with the re-ordering? I can try and rollback  .

No don't bother, doing this is gonna be a pain too. I had a quick look without any diff too and it looks good.

gulpfile.js
wouter-admiraal-sonarsource: No. You have to pass `parallel()` or `series()`. Again, it's because they're deprecating `task()`, but they probably won't remove it before Gulp 5. So we have a few years ahead of us 🙂 .

PR #108, opened 2020-01-15
Save BuildId when preparing

agehrke:
The prepare-task saves a `projectVersion` passed to the task a input parameter. This value is shown in SonarQube UI and thus should be human readable. 

I assume the normal flow is to run prepare and scan tasks for each build of the source code. I therefor think it makes sense to save Azure DevOps `Build.BuildId` as part of the prepare-task. This id could be used to uniquely identify the "scan" when using SonarQube "project_analyses" API.

### DIFF
--- a/common/ts/prepare-task.ts
+++ b/common/ts/prepare-task.ts
@@ -44,6 +44,9 @@ export default async function prepareTask(endpoint: Endpoint, rootPath: string) 
     .map(keyValue => keyValue.split(/=(.+)/))
     .forEach(([k, v]) => (props[k] = v));
 
+  // Save BuildId so that the scan can later uniquely be identified in "project_analyses API" or similar
+  props['sonar.buildid'] = tl.getVariable('Build.BuildId');
+  
   tl.setVariable('SONARQUBE_SCANNER_MODE', scannerMode);
   tl.setVariable('SONARQUBE_ENDPOINT', endpoint.toJson(), true);
   tl.setVariable(


common/ts/prepare-task.ts
```diff
@@ -44,6 +44,9 @@ export default async function prepareTask(endpoint: Endpoint, rootPath: string)
     .map(keyValue => keyValue.split(/=(.+)/))
     .forEach(([k, v]) => (props[k] = v));
 
+  // Save BuildId so that the scan can later uniquely be identified in "project_analyses API" or similar
+  props['sonar.buildid'] = tl.getVariable('Build.BuildId');
```
agehrke: Note that I have no idea what key the buildid should be saved to. In this PR I just assume the key would be called `sonar.buildid`.

mickael-caro-sonarsource:
Hi, 

You can totally do something like this in the project version field : 

4.0.0.$(Build.BuildId) , then the buildid will be determined and replaced at runtime for a full version w/ that buildId. 

Mickaël

agehrke:
Yeah, but I think it makes sense to align build names in SonarQube and ADO, and for the build names to be human understandable. I would prefer not having an internal ADO buildid shown in build names.

mickael-caro-sonarsource:
What is the build name for you in SonarQube ? Where should it appear ?

agehrke:
I think you call it “tags” or version in some parts of SonarQube UI.

mickael-caro-sonarsource:
Ok. Then i'm affraid that your PR is not enough, there's no current way to add custom key/value variables in order to retrieve them in the UI afterwards.

I suggest that you propose a new feature in our community forum here : https://community.sonarsource.com/c/suggestions/features/14

We'll decide what we can do with that.

Thanks anyway for your contribution, i'm closing this PR for now.

agehrke:
No no, projectVersion passed to prepare-task is shown in UI. I want to keep that as is. All I propose is that the prepare-task saves the buildid from ADO so that it can fetched via SonarQube API, eg project_analyses.

agehrke:
Here's example response for `GET api/project_analyses/search` as documented in SonarQube Enterprise EditionVersion 7.9.1 (build 27448):

```json
{
  "analyses": [
    {
      "key": "A1",
      "date": "2016-12-11T17:12:45+0100",
      "projectVersion": "1.2",
      "buildString": "1.2.0.322",
      "revision": "bfe36592eb7f9f2708b5d358b5b5f33ed535c8cf",
      "manualNewCodePeriodBaseline": true,
      "events": [
        {
          "key": "E11",
          "category": "QUALITY_GATE",
          "name": "Quality Gate is Red (was Orange)",
          "description": "Coverage is \u003c 80%"
        },
        {
          "key": "E12",
          "category": "VERSION",
          "name": "6.3"
        }
      ]
    }
]
```

There's no documentation of the properties, but I expect the `projectVersion` is from the `projectVersion` passed in prepare-task.

jacek-poreda-sonarsource:
Hello @agehrke, 

PR in this form is not working the way you would like to. Also we don't plan to save any custom data specific to VSTS vendor for now. As suggested by @mickael-caro-sonarsource please consider creating that topic on community forum.

PR #135, opened 2020-11-02
VSTS-242 - Bumped SonarScanner for .NET version 5.0.0 + extension versions.

mickael-caro-sonarsource:

### DIFF
--- a/extensions/sonarqube/tasks/analyze/old/task.json
+++ b/extensions/sonarqube/tasks/analyze/old/task.json
@@ -9,7 +9,7 @@
   "author": "SonarSource",
   "version": {
     "Major": 3,
-    "Minor": 11,
+    "Minor": 12,
     "Patch": 0
   },
   "demands": ["msbuild", "java"],
--- a/config/config.js
+++ b/config/config.js
@@ -1,4 +1,4 @@
-const msBuildVersion = "4.10.0.19059";
+const msBuildVersion = "5.0.0.23533";
 const cliVersion = "4.4.0.2170"; // Has to be the same version as the one embedded in the Scanner for MSBuild
 
 const scannerUrlCommon =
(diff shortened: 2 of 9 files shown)


extensions/sonarqube/tasks/analyze/old/task.json
```diff
@@ -9,7 +9,7 @@
   "author": "SonarSource",
   "version": {
     "Major": 3,
-    "Minor": 11,
+    "Minor": 12,
```
duncanp-sonar, line 12: The S4MSB release was a major version change because it contains breaking changes, but this is only a minor version update in the Azdo extension. What impact is this going to have on existing users who are using older versions of SQ?

mickael-caro-sonarsource: We still keep the old compatibility with SQ 6.7, i don't think that there are breaking changes here?

duncanp-sonar: I thought the previous version of the scanner was compatible with versions older than SQ 6.7, and the scanner change was a major version change because it did have breaking changes. Is that not the case?

mickael-caro-sonarsource: I always seen LTS 6.7 since i arrived on the docs, that's the reason why. I don't have the whole history

duncanp-sonar: FYI I asked JulienH about scanner compatibility this morning. Historically, apparently there wasn't an official policy. The informal policy was to avoid breaking changes if possible because of the pain this caused users in updating their CI machines. There doesn't appear to have been an assumption that the compatibility policy was the same as for the core SonarQube product. I haven't seen anything in the scanner/Azdo documentation about the minimum supported SonarQube versions so it's difficult to know what users expectations are.

For me, there's a difference between a breaking change in a new version that users have to opt in to, and a breaking change that is pushed on them and for which they don't have a workaround. My concern is that this change falls in to the latter category. On the plus side, SQ v6.7 is quite old, and maybe no users will be affected. The important thing is to assess the risk and that any likely fallout from affected users is acceptable, probably in discussion with the PM.

It would also be worth clarifying what the scanner and Azdo extension versioning/compatibility policy is and make sure it is clearly documented.

PR #145, opened 2021-02-19
Support branch detection for GitHub Enterprise

mingyaulee:
Adds support for branch detection in the prepare analysis task when the repository provider is GitHub Enterprise.

### DIFF
--- a/common/ts/prepare-task.ts
+++ b/common/ts/prepare-task.ts
@@ -95,7 +95,7 @@ export async function populateBranchAndPrProps(props: { [key: string]: string })
     const currentBranch = tl.getVariable("Build.SourceBranch");
     if (provider === "TfsGit") {
       isDefaultBranch = currentBranch === (await getDefaultBranch(collectionUrl));
-    } else if (provider === "Git" || provider === "GitHub") {
+    } else if (provider === "Git" || provider === "GitHub" || provider === "GitHubEnterprise") {
       // TODO for GitHub we should get the default branch configured on the repo
       isDefaultBranch = currentBranch === "refs/heads/master";
     } else if (provider === "Bitbucket") {


pierre-guillot-sonarsource:
Hey @mickael-caro-sonarsource, any opinion?

mickael-caro-sonarsource:
Hi @mingyaulee 

No need to do that, GitHub Enteprise is already well supported for SonarQube (provider remains GitHub). 

Any issue with your configuration maybe ? Please open an issue, i'm closing this PR. 

Thanks.

mingyaulee:
Hi @mickael-caro-sonarsource, if it is well supported then why is [this line](https://github.com/mingyaulee/sonar-scanner-vsts/blob/f1951e682607eec263bf045b63686841eec69c5b/common/ts/prepare-task.ts#L83) checking for `provider === "GitHubEnterprise"`?
I am using GitHub Enterprise with Azure DevOps build pipeline and the branch name is not set for `sonar.branch.name` at all.

mickael-caro-sonarsource:
And my bad for that, i read too fast.

shansmec:
New version not working for Community edition. Its by default passing the branch name to sonarqube since community edition not supports the branch wise analysis. so we are not able to publish the report to sonarqube using GitHub enterprise server. Can any one help me on this.

mingyaulee:
Maybe another configuration needs to be added as task input, i.e. SonarQube edition with options Enterprise and Community. When running the prepare task, if the edition is Community then do not set the sonar.branch.name property.

Either way, this should be reported as a new bug if it does not exist yet because based on the codes it affects all providers when used with SonarQube Community edition, not just GitHub Enterprise.

@mickael-caro-sonarsource what do you think?

brandonaut:
This PR breaks my team's ability to specify our own default branch, by trying to automatically (but incompletely) set the branch parameters. We use GHE with a default branch of `main`, and have our pipelines decide which branch parameters to pass to SQ based on the default branch and PR status. 

This PR forces us to use a default branch of `master`, otherwise the task passes the `sonar.branch.name` parameter to SQ, thereby making it impossible for us to signify if something else is the default branch.

IMO the best solution is to implement the TODO and get the actual default branch from GitHub, instead of assuming it's `master`. It looks like the GitHub REST API could be used, since it provides a `default_branch` field: https://docs.github.com/en/rest/reference/repos#get-a-repository.

PR #183, opened 2022-02-08
VSTS-270 Revert Scanner for .NET to 5.4.1.41282

mickael-caro-sonarsource:

### DIFF
--- a/extensions/sonarqube/vss-extension.json
+++ b/extensions/sonarqube/vss-extension.json
@@ -2,7 +2,7 @@
   "manifestVersion": 1,
   "id": "sonarqube",
   "name": "SonarQube",
-  "version": "5.3.0",
+  "version": "5.4.0",
   "branding": {
     "color": "rgb(67, 157, 210)",
     "theme": "dark"
--- a/extensions/sonarqube/vss-extension.json
+++ b/extensions/sonarqube/vss-extension.json
@@ -2,7 +2,7 @@
   "manifestVersion": 1,
   "id": "sonarqube",
   "name": "SonarQube",
-  "version": "5.4.0",
+  "version": "5.3.1",
   "branding": {
     "color": "rgb(67, 157, 210)",
     "theme": "dark"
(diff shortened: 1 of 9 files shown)


extensions/sonarqube/vss-extension.json
```diff
@@ -2,7 +2,7 @@
   "manifestVersion": 1,
   "id": "sonarqube",
   "name": "SonarQube",
-  "version": "5.3.0",
```
andrei-epure-sonarsource, line 5: Shouldn't this be a bugfix release?

csaba-sagi-sonarsource (approved):
LGTM!

extensions/sonarqube/vss-extension.json
csaba-sagi-sonarsource: I think the bugfix release will be the one once we really fix the scanner, not?

csaba-sagi-sonarsource: BTW If we wait like an hour or to, the scanner bugfix release might be done and we could embed that version of the scanner right away?

mickael-caro-sonarsource: Yes you're right, i'm fixing it.

andrei-epure-sonarsource: @csaba-sagi-sonarsource I would rather mitigate the issue ASAP and we can do the bugfix release of the scanner also ASAP but this is affecting users as we speak so 1h matters

andrei-epure-sonarsource (approved):
Lgtm

extensions/sonarqube/vss-extension.json
mickael-caro-sonarsource: We just have to think of the cache that Azure Pipelines apply on the extensions : if there are too many release in a short period of time, it takes time before it's actually rolled out everywhere.

andrei-epure-sonarsource: And many users cannot downgrade so are blocked

andrei-epure-sonarsource: There is no hurry for the AZDO 2nd release, and it should only be done in a morning (we need a postportem for this).

I meant that we need to do the bugfix release for the S4NET , for the people manually using the S4NET in their pipelines.

PR #186, opened 2022-02-10
VSTS-203 Add Integration tests

mickael-caro-sonarsource:


mickael-caro-sonarsource:
@andrei-epure-sonarsource i crafted here the very first baby step towards having good integrations tests for the Azure DevOps extension. 

It will : 
- Build and package the extension
- Publish it
- Install it into a specific azure devops organization for ITs
- Run those ITs (for now only one per type of scanning : CLI, MSBuild/.NET, maven and gradle) and assert some value around task versions, ncloc and coverage.

Since the run IT part is written in c# (as well as bringing a fresh azure pipeline yaml file), would you mind you or someone from the .NET bubble having a look at those ? That would be helpful for me. 

Thanks in advance.

andrei-epure-sonarsource:
Thanks @mickael-caro-sonarsource , added also @duncanp-sonar and @pavel-mikula-sonarsource to have a look.

its/Models/MeasureModel.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
```
andrei-epure-sonarsource, line 1: pls add header comments in CS files

its/SonarCloudCallWrapper.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace AzureDevOpsExtension.IntegrationTests
+{
+    public class SonarCloudCallWrapper
+    {
+        private string GetBase64EncodedToken()
+        {
+            return Convert.ToBase64String(ASCIIEncoding.ASCII.GetBytes(String.Format("{0}:", Environment.GetEnvironmentVariable("SC_TOKEN"))));
```
andrei-epure-sonarsource, line 16: I don't fully understand why using String.Format, can it be replaced with an interpolated string?

its/SonarCloudIntegrationTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        [TestMethod]
+        public async Task Execute_Gradle_Build_And_Analysis()
+        {
+            Debug.WriteLine("[Gradle]Deleting SonarCloud project if it already exists...");
+            //We first delete the project on SonarCloud
+            var isProjectDeleted = await _scInstance.DeleteProjectAsync(Environment.GetEnvironmentVariable("SC_GRADLE_PROJECT_KEY"));
+
+            Assert.AreEqual(true, isProjectDeleted);
```
andrei-epure-sonarsource, line 62: can be `Assert.IsTrue`

its/SonarCloudIntegrationTests.cs
```diff
@@ -0,0 +1,192 @@
...
+            Assert.AreEqual(BuildResult.Succeeded, currentBuildResult.Result);
+
+            await AssertVersionsOfTasksAreCorrect(currentBuildResult);
+
+            await AssertNcLocAndCoverage(Environment.GetEnvironmentVariable("SC_DOTNET_PROJECT_KEY"), 43, 25);
+        }
+
+        private static VssConnection GetAzDoConnection()
+        {
+            VssBasicCredential credentials = new VssBasicCredential(string.Empty, Environment.GetEnvironmentVariable("AZURE_TOKEN"));
```
andrei-epure-sonarsource, line 124: I think you can move the `Environment.GetEnvironmentVariable` calls into static final variables to improve readability

andrei-epure-sonarsource (commented):
intermediate comments

its/SonarCloudIntegrationTests.cs
```diff
@@ -0,0 +1,192 @@
...
+            Assert.AreEqual(expectedPublishVersion, actualPublishVersion.Value.ToString());
+                
+        }
+
+        private async Task<Build> ExecuteBuildAndWaitForCompleted(string pipelineName)
+        {
+            var definitions = await _buildHttpClient.GetDefinitionsAsync(project: Environment.GetEnvironmentVariable("AZDO_ITS_PROJECT_NAME"));
+            var target = definitions.FirstOrDefault(d => d.Name == pipelineName);
+
+            var queuedBuild = await _buildHttpClient.QueueBuildAsync(new Build
```
andrei-epure-sonarsource, line 168: where are the projects that are being built?

mickael-caro-sonarsource: sample repository in sonarsource-integration-tests organization on AzDo

azure-pipelines.yml
```diff
@@ -79,18 +81,30 @@ stages:
...
+
+            $version = Get-AzureDevopsTaskVersion "$env:BUILD_SOURCESDIRECTORY/build/extensions" "sonarcloud" "publish" "v1"
+            Write-Host "##vso[task.setvariable variable=SC_V1_PUBLISH_VERSION;isOutput=true]$version"
         pwsh: true
     - task: PublishBuildArtifacts@1
       displayName: 'Publish artifacts on the build'
       inputs:
         PathtoPublish: '$(Agent.BuildDirectory)/a'
         ArtifactName: 'extensions'
 - stage: unpublish
```
duncanp-sonar, line 99: Presumably this is un-publishing the "test" version of the artefact? If so, it would be clearer if the stage name/display name made it clear it's referring to the test version.
Same of the "publish" stage below.

azure-pipelines.yml
```diff
@@ -79,18 +81,30 @@ stages:
...
+  condition: and(succeeded(), ne(variables['Build.SourceBranch'], 'refs/heads/master'))
```
duncanp-sonar, line 100: 1) Why not run the tests on master too?

2) the "Build.SourceBranch" condition is repeated three times in the yaml. It would be easier to understand and maintain if it the value was asigned to a variable with more friendly name e.g. `runITs`

commonv5/ts/analyze-task.ts
```diff
@@ -1,12 +1,19 @@
...
 import Scanner, { ScannerMode } from "./sonarqube/Scanner";
 import JavaVersionResolver from "./helpers/java-version-resolver";
+import { extractAndPublishTaskVersion } from "./helpers/task-version-utils";
 import { EndpointType, EndpointData } from "./sonarqube/Endpoint";
 import { sanitizeVariable, PROP_NAMES } from "./helpers/utils";
 
 const JAVA_11_PATH_ENV_NAME = "JAVA_HOME_11_X64";
 
 export default async function analyzeTask(rootPath: string, isSonarCloud: boolean = false) {
+  if (tl.getVariable("IT_TRIGGER_BUILD")) {
```
duncanp-sonar, line 12: I'm not clear what's happening here. It looks like there is test-specific behaviour hard-coded into the "real" tasks to make them export extra data if they detect the ITs are being run. The same data is then passed to the ITs, which check the values are the same.

What is the overall purpose of these checks?
Is it to check that ITs are testing the correct tasks i.e. that the new tasks have been correctly published and installed in the test org?

Or is to check that the published tasks have the expected version numbers? If so, isn't the data just being extracted from the same `.json` files in two different places, once in the product tasks and once in the yaml file? Since the ultimate source of the data is the same in each case, I'm don't understand how they would ever be different.

its/SonarCloudIntegrationTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            _scInstance = new SonarCloudCallWrapper();
+            _buildParameters = "{\"IT_TRIGGER_BUILD\":\"true\"}";
+            VssConnection connection = GetAzDoConnection();
+
+            _buildHttpClient = connection.GetClient<BuildHttpClient>();
+        }
+
+        [TestMethod]
+        public async Task Execute_Scannercli_Build_And_Analysis()
```
duncanp-sonar, line 34: It looks like the four tests are identical except for four variable (pipeline to run, project key, expected LOC, expected coverage). 

Also, the configuration is quite complicated: the loc/coverage numbers are hard-coded, but the pipeline/project key are pipeline variables, passed through the yaml file and picked up by the tests as environment variables.

Could the duplication be removed and the setup simplified by making the tests data-driven? e.g.

`[TestMethod]`
`[DataRow("cli-pipeline", "cli-proj", 123, 456)]`
...

`public async Task Execute_Build_Analysis(string pipelineName, string projectKey, int expectedLoc, double expectedCoverage)`
...

its/SonarCloudIntegrationTests.cs
duncanp-sonar: Could you document the overall structure of the test environment please? i.e. what the various organisations and pipelines are used for. It's hard to understand the big picture just by looking at the yaml/code.

duncanp-sonar (commented):
Thanks for doing this @mickael-caro-sonarsource.

A few comments/questions inline...

azure-pipelines.yml
mickael-caro-sonarsource: Yes is it, ok noted.

azure-pipelines.yml
mickael-caro-sonarsource: I thought the whole process is not necessary on master, but we can run it for sure, i'll remove the condition.

commonv5/ts/analyze-task.ts
mickael-caro-sonarsource: I wanted to make sure that the extension i published during the QA is the same those pipeline/projects used to run, i don't like also the way it is but i haven't found other way to do so.

duncanp-sonar: Can we detect if the unpublish/republish steps fail? That might be enough.

its/SonarCloudIntegrationTests.cs
mickael-caro-sonarsource: @andrei-epure-sonarsource @duncanp-sonar i did some change to centralize and have some dedicated classes to manage those values; should be easier to read now.

its/SonarCloudIntegrationTests.cs
mickael-caro-sonarsource: Here's the doc : https://xtranet-sonarsource.atlassian.net/wiki/spaces/SSG/pages/2298675258/Integration+Tests

azure-pipelines.yml
mickael-caro-sonarsource: Names have been modified

azure-pipelines.yml
mickael-caro-sonarsource: Condition removed

commonv5/ts/analyze-task.ts
mickael-caro-sonarsource: All steps are conditioned to succeeded previous step, is it enough then ?

mickael-caro-sonarsource:
@duncanp-sonar i finished applying your review i think, let me know what you think now. Thanks !

commonv5/ts/analyze-task.ts
duncanp-sonar: I would have thought that would be enough; if the unpublish/publish works, we should be able to assume that the correct version is installed.

mickael-caro-sonarsource: Yes, thanks, however i had a thought yesterday and this morning that this mechanism shall not work if multiple PRs are opened in the same time, because the same version (or a newer) can be installed and replace the one that a PR has installed itself, for the moment since i'm working alone on that it should be ok, but will not be sustanaible and will have to think of being possible to have multiple extension (with revision on the semver and possibly a different id) installed at the same time on one publisher.

duncanp-sonar: Good point. Maybe create a ticket to address that later?

mickael-caro-sonarsource: Yep !

mickael-caro-sonarsource: Ticket created : https://jira.sonarsource.com/browse/VSTS-276

its/SonarCloudCallWrapper.cs
```diff
@@ -0,0 +1,135 @@
...
+                var httpClient = GetHttpClient();
+
+                var httpResult = await httpClient.GetAsync($"/api/measures/component?component={component}&metricKeys=coverage,ncloc");
+
+                var resultContent = JsonConvert.DeserializeObject<MeasureBaseModel>(await httpResult.Content.ReadAsStringAsync());
+
+                return resultContent;
+
+            }
+            catch (Exception ex)
```
duncanp-sonar, line 67: I'm not sure catching and squashing the server exceptions is helpful; the test is still going to fail later (e.g. an assertion fails), but it will take more digging to work out why.
Test execution failures will be easier to diagnose if we let the test framework catch and log the underlying exception.

duncanp-sonar (approved):
Looks good. Just a couple of suggestions to make diagnosing test failures easier in the future.

its/SonarCloudCallWrapper.cs
mickael-caro-sonarsource: Agreed, i'll remove those catches

its/SonarCloudIntegrationTests.cs
```diff
@@ -0,0 +1,136 @@
...
+namespace AzureDevOpsExtension.IntegrationTests
+{
+    [TestFixture]
+    public class SonarCloudIntegrationTests
+    {
+
+        private SonarCloudCallWrapper _scInstance;
+        private BuildHttpClient _buildHttpClient;
+
+        private static string _azureToken => Environment.GetEnvironmentVariable("AZURE_TOKEN");
```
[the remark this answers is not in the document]
mickael-caro-sonarsource: Makes sense, i added that as well, thanks !

its/Models/DotnetFrameworkTestCase.cs
```diff
@@ -0,0 +1,36 @@
...
+	public class DotnetFrameworkTestCase : BuildTestCase
+	{
+		public DotnetFrameworkTestCase()
+		{
+			base.Coverage = Double.Parse(EnvironmentVariableWrapper.GetVariableOrThrow("ITS_DOTNET_EXPECTED_COVERAGE"));
+			base.NcLocs = Int64.Parse(EnvironmentVariableWrapper.GetVariableOrThrow("ITS_DOTNET_EXPECTED_NCLOCS"));
+			base.PipelineName = EnvironmentVariableWrapper.GetVariableOrThrow("ITS_DOTNET_PIPELINE_NAME");
+			base.ProjectKey = EnvironmentVariableWrapper.GetVariableOrThrow("ITS_DOTNET_SC_PROJECT_KEY");
+			base.LogPrefix = "S4NET - .NET Framework";
+		}
```
pavel-mikula-sonarsource, line 33: You can save a ton of duplications if you move the initialization to the base class and provide just the environment variable infix.

It will also enforce regular structure for the environment variable names.
```suggestion
		public DotnetFrameworkTestCase() : base("DOTNET", "S4NET - .NET Framework") {}
```
For sake of simplicity, we can use the infix as the LogPrefix as well.
```suggestion
		public DotnetFrameworkTestCase() : base("DOTNET") {}
```

its/Models/CobolTestCase.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace IntegrationTests.Models
+{
+	public class CobolTestCase : BuildTestCase
+	{
+		public CobolTestCase()
+		{
+			base.Coverage = Double.Parse(EnvironmentVariableWrapper.GetVariableOrThrow("ITS_SCANNERCLI_EXPECTED_COVERAGE"));
+			base.NcLocs = Int64.Parse(EnvironmentVariableWrapper.GetVariableOrThrow("ITS_SCANNERCLI_EXPECTED_NCLOCS"));
+			base.PipelineName = EnvironmentVariableWrapper.GetVariableOrThrow("ITS_SCANNERCLI_PIPELINE_NAME");
+			base.ProjectKey = EnvironmentVariableWrapper.GetVariableOrThrow("ITS_SCANNERCLI_SC_PROJECT_KEY");
```
pavel-mikula-sonarsource, line 31: Environment variables are called `SCANNERCLI` and the test case is called `Cobol`. That's inconsistent and confusing

pavel-mikula-sonarsource:
@mickael-caro-sonarsource I just saw I was late for the party. Will you create a cleanup PR?

its/Models/DotnetFrameworkTestCase.cs
pavel-mikula-sonarsource: Having that, we can remove all 4 of these classes and initialize the value only in the tests.

PR #257, opened 2023-06-07
Fix deprecated SQ extension cateogry

7PH:
These are the categories that are accepted by Azure Marketplace: https://learn.microsoft.com/en-us/azure/devops/extend/develop/manifest?view=azure-devops#required-attributes

When trying to upload a test build of the extension, here is the error shown:
image


7PH:
Closing as won't fix (for now), should be part of a later fix which ensures compatibility with v4 extension

PR #266, opened 2023-06-14
VSTS-306 Fallback to default report task location when publish task can not find report task variable

7PH:


commonv5/ts/sonarqube/TaskReport.ts
```diff
@@ -53,11 +73,8 @@ export default class TaskReport {
...
       taskReportGlobResult = tl.findMatch(tl.getVariable("Agent.BuildDirectory"), taskReportGlob);
     } else {
-      const reportTaskFile = tl.getVariable("SONARQUBE_SCANNER_REPORTTASKFILE");
-      if (!reportTaskFile) {
-        throw TaskReport.throwInvalidReport("[SQ] Unable to find report task file.");
-      }
-      taskReportGlob = reportTaskFile;
+      taskReportGlob =
+        tl.getVariable("SONARQUBE_SCANNER_REPORTTASKFILE") || TaskReport.getDefaultPathGlob();
       taskReportGlobResult = tl.find(taskReportGlob);
```
mickael-caro-sonarsource, line 78: I think you need to use tl.findMatch here instead of find in case we have the ** pattern on it.

mickael-caro-sonarsource (approved):

PR #291, opened 2023-09-21
update azure pipeline library

quentin-chevrin-sonarsource:
Please review our [contribution guidelines](https://github.com/SonarSource/sonar-scanner-vsts/blob/master/docs/contributing.md).


commonv5/ts/helpers/__tests__/request-test.ts
```diff
@@ -1,35 +0,0 @@
-import Endpoint, { EndpointType } from "../../sonarqube/Endpoint";
-import { getJSON } from "../request";
-
-const mockCallback = jest.fn((_options, cb) => cb(null, { name: "a" }, null));
-
-jest.mock("request", () => ({
-  get: (options, cb) => mockCallback(options, cb),
-}));
-
-describe("request", () => {
```
antonio-garcia-sonarsource, line 10: Can't we keep this test?

commonv5/ts/helpers/request.ts
```diff
@@ -7,60 +7,40 @@ interface RequestData {
...
+      } else {
+        tl.debug(`[SQ] API GET '${path}' failed, error is ${error.message}`);
+      }
+    });
 }
 
 export function getJSON(endpoint: Endpoint, path: string, query?: RequestData): Promise<any> {
-  return get(endpoint, path, true, query);
+  return get(endpoint, path, query);
 }
```
antonio-garcia-sonarsource, line 42: Is this necessary? It looks like it's not adding anything.

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     /* Basic Options */
     "target": "ES5",
     "module": "commonjs", 
-    "lib": [ "es2017" ],    
+    "lib": [ "es2017", "DOM" ],
```
antonio-garcia-sonarsource, line 6: Are you using `window` or `document`?

commonv5/ts/helpers/__tests__/request-test.ts
quentin-chevrin-sonarsource: This is not testing anything. It tests that the "timeout" property is set, which is set with no condition in the code.

And with axios it's super hard to test that.

tsconfig.json
quentin-chevrin-sonarsource: Axios is

antonio-garcia-sonarsource (approved):
Code LGTM! But I couldn't do proper functional validation; I let the others with more knowledge of the plugin do that.

common/ts/helpers/request.ts
```diff
@@ -7,59 +7,33 @@ interface RequestData {
...
-export function getJSON(endpoint: Endpoint, path: string, query?: RequestData): Promise<any> {
-  return get(endpoint, path, true, query);
+      return response.data;
+    })
+    .catch((error) => {
+      if (error.response) {
+        tl.debug(`[SQ] API GET '${path}' failed, status code was: ${error.response.status}`);
+      } else {
+        tl.debug(`[SQ] API GET '${path}' failed, error is ${error.message}`);
+      }
```
stanislavhh, line 33: You dropped `logAndReject` fn which makes sense to me ( I don't see any benefit of it), but I think we should throw an error here too.  Otherwise any task will continue to work without the response we should have received, if I understand correctly.

stanislavhh (commented):
I did FV of SQ extension with: 
-maven ✅ 
-cfamily ✅ 
-other ✅ 

Nicely done with axios, I have one minor comment. other looks good to me

common/ts/package.json
```diff
@@ -5,12 +5,11 @@
   "author": "SonarSource",
   "license": "LGPL-3.0",
   "dependencies": {
-    "azure-devops-node-api": "^9.0.1",
-    "azure-pipelines-task-lib": "^2.8.0",
+    "axios": "0.27.2",
```
7PH, line 8: Why use axios version `0.27.2` rather than a more recent one? [Latest seems to be 1.5.0](https://www.npmjs.com/package/axios)

7PH (approved):
FV ok with gradle and 'other CI' on SQ LTS and latest ✅

common/ts/helpers/request.ts
quentin-chevrin-sonarsource: Yes I was a bit undecided about that, because I saw the error was not handled and the reject was giving the same information as the log. But I'm fine either way, happy to throw something to be safe

common/ts/package.json
quentin-chevrin-sonarsource: I tried to use the latest axios version, the problem is that they use modern esmodule syntax in their library and the build system of this project is not fit for that. I had a lot of errors because of that, and I tried to fix them, but I was going down a big rabbit hole and I felt it was too much work.

quentin-chevrin-sonarsource: https://stackoverflow.com/questions/73958968/cannot-use-import-statement-outside-a-module-with-axios

This link gives more info about that

antonio-garcia-sonarsource: I hope it doesn't introduce a different set of vulnerabilities 😅

package.json
```diff
@@ -39,18 +40,16 @@
     "needle": "3.2.0",
     "openpgp": "5.9.0",
     "prettier": "2.8.8",
-    "request": "2.88.2",
```
quentin-chevrin-sonarsource, line 42: Are you sure about this one ? request is still used in gulpfile.js no ? @stanislavhh

mickael-caro-sonarsource:
@quentin-chevrin-sonarsource @7PH I see you updated some stuff in the common/ts folder : be careful with this one as it will target the V4 of SQ if you bump those versions : we still need to keep backward compatibility with older versions and given the breaking changes the azure devops node API bump, i am not sure this is something we want. I can only advise to bump the versions on the code but not release the V4 (it should be, I think, deprecated anyway at some point).

PR #294, opened 2023-10-10
[NO-JIRA] Refactor build process

7PH:
Minor code reorganizations and comments to ease understanding of what is happening during the build process, including with downloading msbuild scanners and extracting cli scanner. There _should not_ be any functional changes.

- Moved utilities out of gulpfile into config/utils
- Renamed some misleading gulp task names
- Added comments around code that downloads and extract scanners & reorder code semantically (per scanner kind)


stanislavhh (approved):
Looks good ✅

PR #299, opened 2023-10-18
[NO-JIRA] Fix axios sending custom user agent to SonarQube instance

7PH:
As part of a recent [dependency upgrade sprint](https://github.com/SonarSource/sonar-scanner-vsts/pull/291), the obsolete request library was changed to axios. 

Axios is sending, by default, is sending its user agent to SonarQube instance. While this is not an issue for SQ, some specific and custom setups with reverse proxy may reject requests from Azure CI before they reach the SQ instance, based on this HTTP header  `User-Agent: axios/0.27.2`.

This PR resets this user agent so that it is not sent (corresponding to the behavior before `5.16.0`).


stanislavhh (approved):
LGTM

PR #306, opened 2023-11-09
Fix ESLint & Prettier configuration

7PH:
Currently, we have to drop auto-formatting in the extension when we work on it. Also, there are some ESLint false positive errors.
This PR focuses on developer experience and formats all the code with `npm run format` properly

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -119,13 +119,11 @@
     "import/no-deprecated": "error",
     "import/no-duplicates": "error",
     "import/no-dynamic-require": "error",
-    "import/no-extraneous-dependencies": ["error", { "packageDir": ["./", "./common/ts"] }],
     "import/no-mutable-exports": "error",
     "import/no-named-as-default": "error",
     "import/no-named-as-default-member": "error",
     "import/no-named-default": "error",
     "import/no-webpack-loader-syntax": "error",
-    "import/order": ["error", { "groups": ["builtin", "external"], "newlines-between": "never" }],
 
     // does not properly work with ts
     "import/no-unresolved": "off",
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "merge-stream": "2.0.0",
     "needle": "3.2.0",
     "openpgp": "5.10.1",
-    "prettier": "2.8.8",
+    "prettier": "3.0.3",
     "request": "2.88.2",
     "semver": "5.5.0",
     "sonarqube-scanner": "3.0.0",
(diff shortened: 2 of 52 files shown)


.eslintrc
```diff
@@ -119,13 +119,11 @@
     "import/no-deprecated": "error",
     "import/no-duplicates": "error",
     "import/no-dynamic-require": "error",
-    "import/no-extraneous-dependencies": ["error", { "packageDir": ["./", "./common/ts"] }],
```
7PH, line 122: Not necessary, because TS already throws an error when a dependency types are not resolved 

Currently produces such false positives
image

package.json
```diff
@@ -39,7 +39,7 @@
     "merge-stream": "2.0.0",
     "needle": "3.2.0",
     "openpgp": "5.10.1",
-    "prettier": "2.8.8",
+    "prettier": "^3.0.3",
```
7PH, line 42: Update prettier to v3

tsconfig.json
```diff
@@ -13,6 +13,7 @@
   },
   "include": [
     "./common/ts/**/*.ts",
+    "./commonv5/ts/**/*.ts",
```
7PH, line 16: Was causing false positive type errors in `commonv5` libraries

.eslintrc
```diff
@@ -119,13 +119,11 @@
...
     "import/no-mutable-exports": "error",
     "import/no-named-as-default": "error",
     "import/no-named-as-default-member": "error",
     "import/no-named-default": "error",
     "import/no-webpack-loader-syntax": "error",
-    "import/order": ["error", { "groups": ["builtin", "external"], "newlines-between": "never" }],
```
7PH, line 128: Conflicts with prettier import sort

package.json
7PH: Agree

stanislavhh (approved):
Why do we have to drop auto formatting? Can we add `"trailingComma": "all"` to prettierrc file, seems like it works well.

BTW, great work, I am very happy we'll have better experience after this pr

7PH:
> Why do we have to drop auto formatting?

Before this PR, when I used auto-formatting it created an eslint error in some files

> Can we add "trailingComma": "all" to prettierrc file, seems like it works well

I added it initially, but found out that even without "trailingComma" it adds it, but it does not hurt to add it, I guess, I'll be re adding it 👍

stanislavhh:
> I added it initially, but found out that even without "trailingComma" it adds it, but it does not hurt to add it, I guess, I'll be re adding it 👍

Yeah, it seems that "all" is a default value, but in fact on my side it works differently if I used it explicitly

PR #307, opened 2023-11-09
VSTS-324 Migrate request to axios & Use proxy configuration from Azure

7PH:
This PR includes:
- Migrating to axios library for V5 tasksa
- Using the azure devops library `azure-pipelines-task-lib` to fetch proxy configuration and use it with axios
- Code quality improvements


commonv5/ts/package.json
```diff
@@ -8,8 +8,9 @@
     "azure-devops-node-api": "12.1.0",
     "azure-pipelines-task-lib": "4.4.0",
     "fs-extra": "11.1.1",
-    "request": "2.88.2",
     "guid-typescript": "1.0.9",
+    "node-fetch": "2.7.0",
```
7PH, line 12: Last CJS version

commonv5/ts/package.json
```diff
@@ -8,8 +8,9 @@
...
+    "proxy-agent": "5.0.0",
```
7PH, line 13: Same

commonv5/ts/prepare-task.ts
```diff
@@ -10,16 +10,6 @@ import TaskReport from "./sonarqube/TaskReport";
...
-  if (
-    endpoint.type === EndpointType.SonarQube &&
-    (endpoint.url.startsWith("https://sonarcloud.io") ||
-      endpoint.url.startsWith("https://sonarqube.com"))
-  ) {
-    tl.warning(
-      "There is a dedicated extension for SonarCloud: https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarcloud",
-    );
-  }
-
```
7PH, line 22: I think we can safely remove this annotation, as the split was more than 5 years ago.

commonv5/ts/sonarqube/Endpoint.ts
```diff
@@ -16,26 +18,58 @@ export interface EndpointData {
...
+      this.data.url = this.data?.url.replace(/\/$/, "");
+    }
+    this.proxy = tl.getHttpProxyConfiguration();
   }
 
   public get url() {
     return this.data.url;
   }
 
+  public get auth(): { username: string; password: string } {
```
7PH, line 41: `password` could previously be undefined, that is not the case anymore, it will now be an empty string. That is because that is generally what libraries expect.

eg [axios](https://github.com/axios/axios/blob/7009715369a50740ba2ce00534012c1caf269ad2/index.d.ts#L116) 

If you are setting the auth header by hand, you'll want to use
`Basic username:` (so password as empty string).

commonv5/ts/sonarqube/Endpoint.ts
```diff
@@ -16,26 +18,58 @@ export interface EndpointData {
...
+      )}`,
+    };
+
+    // Add proxy configuration, when relevant
+    if (this.proxy) {
+      options.agent = proxyAgent(this.proxy.proxyFormattedUrl);
+      tl.debug("Using proxy agent: " + JSON.stringify(options.agent));
+    } else {
+      tl.debug("Not using a proxy agent");
+    }
```
7PH, line 68: Uses the agent proxy conf to set proxy for requests

tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "lib": [ "es2017", "DOM" ],    
     "sourceMap": true,
     "noUnusedLocals": true,  
-    "noUnusedParameters": true,    
+    "noUnusedParameters": true,
+    "esModuleInterop": true,
```
7PH, line 10: Needed for `proxy-agent`

gulpfile.js
7PH, line 1: gulpfile wasn't formatted in previous PR, hence all the noise here, I'll point out the changes

gulpfile.js
```diff
@@ -350,451 +367,476 @@ gulp.task('scanner:download', () => {
...
+  return fetch(process.env.ARTIFACTORY_URL + "/api/build", {
+    method: "put",
+    body: JSON.stringify(buildInfo),
+    headers: {
+      "Content-Type": "application/json",
+      Authorization: `Basic ${Buffer.from(
+        `${process.env.ARTIFACTORY_DEPLOY_USERNAME}:${process.env.ARTIFACTORY_DEPLOY_PASSWORD}`,
+      ).toString("base64")}`,
+    },
+  });
```
7PH, line 661: moved to node-fetch

gulpfile.js
```diff
@@ -350,451 +367,476 @@ gulp.task('scanner:download', () => {
...
+      url,
+      buildNumber: process.env.BUILD_NUMBER,
+    }),
+    headers: {
+      "Content-Type": "application/json",
+      Authorization: `Basic ${Buffer.from(
+        `${process.env.BURGR_USERNAME}:${process.env.BURGR_PASSWORD}`,
+      ).toString("base64")}`,
+    },
+  });
```
7PH, line 802: moved to node-fetch

gulpfile.js
```diff
@@ -350,451 +367,476 @@ gulp.task('scanner:download', () => {
...
```
7PH, line 839: moved to node-fetch

commonv5/ts/helpers/request.ts
```diff
@@ -1,68 +1,41 @@
...
-}
-
-export function getServerVersion(endpoint: Endpoint): Promise<semver.SemVer> {
-  return get(endpoint, "/api/server/version", false).then(semver.coerce) as Promise<semver.SemVer>;
+  try {
+    const hasQuery = query && Object.keys(query).length > 0;
+    const response = await fetch(
+      endpoint.url + path + hasQuery ? "?" + new URLSearchParams(query).toString() : "",
+      endpoint.toFetch(),
+    );
```
7PH, line 23: We are losing the ability to use a timeout. We could in theory re-add support with signals and abort controller, but not worth it IMHO

commonv5/ts/helpers/request.ts
```diff
@@ -1,68 +1,41 @@
...
-}
-
-export function getServerVersion(endpoint: Endpoint): Promise<semver.SemVer> {
-  return get(endpoint, "/api/server/version", false).then(semver.coerce) as Promise<semver.SemVer>;
+  try {
+    const hasQuery = query && Object.keys(query).length > 0;
+    const response = await fetch(
+      endpoint.url + path + (hasQuery ? "?" + new URLSearchParams(query).toString() : ""),
+      endpoint.toFetch(),
+    );
```
7PH, line 23: We are losing the ability to use a timeout. We could in theory re-add support with signals and abort controller, but not worth it IMHO

stanislavhh (approved):
LGTM 🤞 

A couple of minor comments

commonv5/ts/helpers/request.ts
```diff
@@ -1,68 +1,41 @@
...
-export function getJSON(endpoint: Endpoint, path: string, query?: RequestData): Promise<any> {
-  return get(endpoint, path, true, query);
-}
-
-export function getServerVersion(endpoint: Endpoint): Promise<semver.SemVer> {
-  return get(endpoint, "/api/server/version", false).then(semver.coerce) as Promise<semver.SemVer>;
+  try {
+    const hasQuery = query && Object.keys(query).length > 0;
+    const response = await fetch(
+      endpoint.url + path + hasQuery ? "?" + new URLSearchParams(query).toString() : "",
```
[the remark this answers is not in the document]
7PH: Yes, it exists and usable in node 10 (but not node 8 or 6)

commonv5/ts/helpers/request.ts
```diff
@@ -1,68 +1,41 @@
...
+      endpoint.toFetch(),
+    );
+    if (isJson) {
+      return await response.json();
+    } else {
+      return await response.text();
+    }
```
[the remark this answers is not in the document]
7PH: > can we get rid of isJson param and use try?

Unfortunately not - we can not read the response twice (`TypeError: body used already`)

We could read `response.contentType` to choose what method to use, but it seems "risky" as it is less flexible than simply knowing what's the expected output format. If an incorrectly configured reverse proxy doesn't specify the correct content type for any request, the pipeline will fail. WDYT?

commonv5/ts/sonarqube/Endpoint.ts
7PH: Agree. We have `toSonarProps` which is similar in the same class which generates a partial scanner property object. This method is very similar as it generates partial node-fetch options, what about `toFetchOptions` to have better meaning but still make it obvious it's different than `toSonarProps` semantically?

commonv5/ts/helpers/request.ts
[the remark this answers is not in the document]
7PH: > NP: To me it was difficult to understand the meaning of toFetch function by its name. maybe something simpler like getOption would work?

Agree. We have `toSonarProps` which is similar in the same class which generates a partial scanner property object. This method is very similar as it generates partial node-fetch options, what about `toFetchOptions` to have better meaning but still make it obvious it's different than `toSonarProps` semantically?

commonv5/ts/helpers/request.ts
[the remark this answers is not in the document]
7PH: > Does node fetch have some default timeout?

Actually after double-checking, there indeed is a way to use a timeout - Thanks for spotting, I'll use the `timeout` option

commonv5/ts/helpers/request.ts
7PH: > Does node fetch have some default timeout?

Actually after double-checking, there indeed is a way to use a timeout - Thanks for spotting, I'll use the `timeout` option

commonv5/ts/helpers/request.ts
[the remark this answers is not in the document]
stanislavhh: Got it, thanks for clarifying

7PH:
Hello @mateustanaka , could you please tell me if `5.18.1` fixed your issue?

PR #310, opened 2023-11-13
VSTS-326 Update jest to use non vulnerable babel/traverse package

stanislavhh:
Only jest and jest-ts libraries were using babel/traverse 7.15. Updating these 2 updated babel/traverse to >7.23 which is not vulnerable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "devDependencies": {
     "@cyclonedx/bom": "^3.8.0",
     "@types/fs-extra": "9.0.13",
-    "@types/jest": "27.0.2",
+    "@types/jest": "29.5.8",
     "@types/jest-when": "2.7.3",
     "@types/node": "14.14.31",
     "@types/request": "2.47.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   "devDependencies": {
     "@cyclonedx/bom": "^3.8.0",
     "@types/fs-extra": "9.0.13",
-    "@types/jest": "27.0.2",
+    "@types/jest": "29.5.8",
```
7PH, line 9: Latest `@types/jest` is `29.6.3` and used jest is `29.7.0` - Could we bump this to at least `29.6.3`?

7PH (approved):
LGTM 👍

package.json
stanislavhh: According to [npm](https://www.npmjs.com/package/@types/jest) the latest is 29.5.8, that's why I use this one.

[GH](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/657a7864cd96fc3159f8c47ac6a1d4d5330a9742/types/jest/package.json#L4C1-L5C1) shows 29.5.999, but where do you see `29.6.3`/

7PH: My bad, was confused with another one 👍

PR #312, opened 2023-11-14
VSTS-221 VSTS-308 Document logic to decide whether we are on the main branch or not & multistage pipeline and major task versions

7PH:
**VSTS-221 Document logic to decide whether we are on the main branch or not**

In the end, I decided to keep the same logic than currently. It is not worth it IMO to take risk to break Developer+ instances to have Community Edition supporting main branch with other names than `master`

The current behavior is:
- For azure repos, we send the property if the analyzed branch is not the default branch. If the Web API can not be reached, the fallback is to send the property only if the analyzed branch is not `master`.
- For SVN, we send the property if the analyzed branch is not `trunk`. 
- For other providers we send the property if and only if the branch is not `master`

This PR focuses on code quality and documentation to make the current behavior clear

____

**VSTS-309 Fail fast if required properties do not exist**

Adds a verbose error message:
- Asking to add the prepare task before the analyze task
- Asking to add the prepare task and analyze task before the publish task
- Specifying that multistage pipelines are not supported (if using multiple stages)
- Specifying that major task versions should not be mixed 

Also added an enum for all the custom variables we use. IMO there was no need to list all the required variables and ensure they were all here at the beginning of the analysis and publish tasks because if one of the required variables is not there, others will not be there as well. I went for [KISS](https://en.wikipedia.org/wiki/KISS_principle)

image

### DIFF
--- a/commonv5/ts/publish-task.ts
+++ b/commonv5/ts/publish-task.ts
@@ -6,21 +6,25 @@ import Endpoint, { EndpointData, EndpointType } from "./sonarqube/Endpoint";
 import Metrics from "./sonarqube/Metrics";
 import Task, { TimeOutReachedError } from "./sonarqube/Task";
 import TaskReport from "./sonarqube/TaskReport";
+import { TASK_MISSING_VARIABLE_ERROR_HINT, TaskVariables } from "./helpers/constants";
 
 let globalQualityGateStatus = "";
 
-export default async function publishTask(endpointType: EndpointType) {
-  const params = tl.getVariable("SONARQUBE_SCANNER_PARAMS");
-  if (!params) {
+export default async function publishTask(_endpointType: EndpointType) {
+  const missingVariables = [
+    TaskVariables.SonarQubeScannerParams,
+    TaskVariables.SonarQubeEndpoint,
+  ].filter((variable) => typeof tl.getVariable(variable) === "undefined");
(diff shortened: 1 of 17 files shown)


commonv5/ts/publish-task.ts
```diff
@@ -6,21 +6,25 @@ import Endpoint, { EndpointData, EndpointType } from "./sonarqube/Endpoint";
...
+export default async function publishTask(_endpointType: EndpointType) {
+  const missingVariables = [
+    TaskVariables.SonarQubeScannerParams,
+    TaskVariables.SonarQubeEndpoint,
+  ].filter((variable) => typeof tl.getVariable(variable) === "undefined");
+  if (missingVariables.length > 0) {
     tl.setResult(
       tl.TaskResult.Failed,
-      `The ${endpointType} Prepare Analysis Configuration must be added.`,
+      `Variables are missing. Please make sure that you are running the Prepare and Analyze tasks before running the Publish task.\n${TASK_MISSING_VARIABLE_ERROR_HINT}`,
```
stanislavhh, line 21: FYI, this will only fail when there is no prepare task. When we skip analysis task there is a warning but publish task succedes. Anyway, IMO it makes sense to have this warning

stanislavhh (approved):
Nice work!
We need to make sure we are waiting for `isDefaultBranch` result

PR #314, opened 2023-11-16
VSTS-323 Document deprecated sonar.pullrequest.provider

stanislavhh:


7PH (approved):
LGTM - Later usages of `sonar.pullrequest.provider` are not prefixed with this comment, but I think it is fine because of the visibility of this 3-line blocks and it's 5-10 lines below

PR #317, opened 2023-11-16
VSTS-324 Drop V4 axios in favor of node-fetch & Fix mend vm2 vulnerability

7PH:
- Completely drop axios in favor of node-fetch in V4 of the extension (will not be released)
- Drop `proxy-agent` in favor of manually instantiating `http-proxy-agent` or `https-proxy-agent` depending on endpoint url


stanislavhh (approved):
LGTM

PR #328, opened 2023-12-20
VSTS-347 Azure PR decoration adopts the new CCT design

7PH:
This PR is split in 3 parts:
- Drop the obsolete `Metric` class and put its logic in a `fetchMetrics` function, covered by tests
- Change styling for build summary to the new design for failing quality gates
- Fixes the master Quality Gate. For some reason, the previous PR was passing coverage but failed master after merge.

Dropped:
- Support build summary when the QG is passing

Known limitations:
- Custom links are not created for complexity reasons: it would need to support SonarCloud, SonarQube (all versions), and would require copying/pasting a lot of code from SonarQube to handle urls
- No icons, only emojis ❌✅


commonv5/ts/sonarqube/Analysis.ts
```diff
@@ -1,58 +1,26 @@
...
-  line-height: normal;
-  font-weight: 500;
-  font-size: 12px;`.trim();
-
-const separatorStyle = `
-  background: #EBECF0;
-  height: 1px;
-  margin: 16px 0px;`.trim();
-
 export default class Analysis {
```
stanislavhh, line 15: I think we could rename this class, as it is too broad considering these changes, wdyt?
Something like HTMLAnalysisReport would be a better fit imo

stanislavhh (approved):
Nice work, code looks much cleaner!

Just a one minor comment, other looks good!

FYI, I have a qualitygate check in my branch policy, and for some reason it's just stuck (don't think it is related to your PR, but let me know if you have any ideas why)
image

stanislavhh:
One thing I see in figma is links to failed conditions. I am curious whether we could add them, and if yes what is the effort in time would that be?
image

commonv5/ts/sonarqube/Analysis.ts
7PH: I 💯% agree, and I think this class doesn't belong to the `sonarqube` folder because it's used for both `SonarQube` and `SonarCloud` - I'll rename it to your suggestion and see if I can move it out of this folder, unless you have concerns

7PH:
> I am curious whether we could add them, and if yes what is the effort in time would that be?

I put some info about this in the PR description:

> Custom links are not created for complexity reasons: it would need to support SonarCloud, SonarQube (all versions), and would require copying/pasting a lot of code from SonarQube to handle urls

So, in theory, it could be done, UX agreed to cut the branch but I'll make it clear that it's technically possible in the future

PR #329, opened 2023-12-21
VSTS-347 Show available measures when they are available through QG conditions even when QG passes

7PH:
The goal with this PR is to as close as possible to the expected result from UX mockup with the data we already have. Only available measures are included in the summary when the QG passes now. More information can be found in our internal VSTS ticket.

image
image
image
image


stanislavhh (approved):
LGTM ✅

PR #331, opened 2024-01-05
VSTS-348 Upgrade azure pipelines task lib to V4.7.0 & Use specific version rather than auto-upgrade syntax

7PH:
The version of azure-pipelines-task-lib used in tasks is fixed to 4.4.0 for SQ V5 tasks and SC tasks, but was set to ^4.4.0 in SQ V4 tasks which means V4.6.0 was used. We should:
- Upgrade to 4.7.0
- Stick to the fixed-version usage and use specifically 4.7.0 for V4 tasks (instead of ^4.7.0)


stanislavhh (approved):
LGTM ✅

PR #332, opened 2024-01-05
VSTS-348 Fix SQ code smell after azure-pipelines-task-lib upgrade

7PH:
After the upgrade of `azure-pipelines-task-lib` to `4.7.0`, 2 code smells were created on Next regarding `ToolRunner.exec(..)` deprecated in favor of `ToolRunner.execAsync(..)`. 

It was not detected in the pull request because it wasn't part of the changed lines

____

Also, this PR fixes the remaining Mend vulnerability that we have on `follow-redirects`. Since we specify already that NPM 8 should be used, it uses `overrides` mechanism from NPM to fix the `follow-redirects` version to the latest.

### DIFF
--- a/common/ts/package-lock.json
+++ b/common/ts/package-lock.json
@@ -10,7 +10,7 @@
       "license": "LGPL-3.0",
       "dependencies": {
         "azure-devops-node-api": "^12.1.0",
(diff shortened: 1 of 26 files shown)


stanislavhh (approved):
Nice work, lgtm ✅

PR #338, opened 2024-01-17
VSTS-350 Fix Next SQ branch analysis and CIRRUS_PR usage

7PH:
[this commit](https://github.com/SonarSource/sonar-scanner-vsts/commit/f7102231e6a0b251fa7bd8f984ed8cb389050cd2) added PR decoration but also introduced a regression for branch builds (1), and it wasn't properly analyzing PRs because of an old bug in the gulp file (2).

(1) This PR fixes the usage of [`CIRRUS_PR`](https://cirrus-ci.org/guide/writing-tasks/#environment-variables) which was never correct. The current code assumes `CIRRUS_PR` can take values `false` or `true` but in fact, these are the actual values:
- On master branch, it's unset
- In PR: image

(2) In PR analyses, no change was detected because the CI environment did not have the `master` branch to compare the changes between the PR branch and the target branch. Adding `git fetch` solves this

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -52,6 +52,7 @@ build_task:
     - npm install
     - npm run install-dep-full
   script:
+    - git fetch
     - source cirrus-env BUILD
     - npm run validate-ci
     - npm run sonarqube
@@ -78,7 +79,7 @@ promote_task:
 mend_scan_task:
   depends_on:
     - build
-  only_if:  $CIRRUS_BRANCH == "master"
+  only_if: $CIRRUS_BRANCH == "master"
   timeout_in: 30m
   eks_container:
     <<: *CONTAINER_TEMPLATE
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -519,11 +519,11 @@ gulp.task("build:test", gulp.series("clean", "copy", "test", "tfx:test"));
  * =========================
  */
 gulp.task("deploy:vsix:sonarqube", () => {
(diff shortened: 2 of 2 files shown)


gulpfile.js
```diff
@@ -519,11 +519,11 @@ gulp.task("build:test", gulp.series("clean", "copy", "test", "tfx:test"));
...
-  if (process.env.CIRRUS_BRANCH !== "master" && process.env.CIRRUS_PR === "false") {
+  if (process.env.CIRRUS_BRANCH !== "master" && ! process.env.CIRRUS_PR) {
     gutil.log("Not on master nor PR, skip deploy:vsix");
     return gutil.noop;
   }
-  if (process.env.CIRRUS_PR === "true" && process.env.DEPLOY_PULL_REQUEST === "false") {
+  if (process.env.CIRRUS_PR && process.env.DEPLOY_PULL_REQUEST === "") {
     gutil.log("On PR, but artifacts should not be deployed, skip deploy:vsix");
     return gutil.noop;
   }
```
7PH, line 529: This code was incorrect, but likely we never encountered this case hence why it was silently existing

.cirrus.yml
```diff
@@ -52,6 +52,7 @@ build_task:
     - npm install
     - npm run install-dep-full
   script:
+    - git fetch
```
7PH, line 55: If we don't run `git fetch`, the master branch does not exist and the scanner can't compare the content of the dev branch (pull request) to the mater branch

stanislavhh (approved):
Nice finding, lgtm ✅

PR #339, opened 2024-01-22
SONAR-21476 Authenticated access to JFrog

antoine-vinot-sonarsource:


azure-pipelines.yml
```diff
@@ -11,8 +11,11 @@ pool:
 
 variables:
   - group: sonar-scanner-vsts-variables
+  - group: artifactory_access
```
7PH, line 14: Loads repox token and URL

azure-pipelines.yml
```diff
@@ -23,22 +26,22 @@ stages:
     variables:
       publisher: $(SonarSourceITPublisher)
     steps:
+    
     - task: Npm@1
       displayName: 'Update npm'
+      env:
+        NPM_CONFIG_//repox.jfrog.io/artifactory/api/npm/:_authToken: $(ARTIFACTORY_PRIVATE_READER_ACCESS_TOKEN)
```
7PH, line 33: Using `npm config set ..` at the top level didn't work, so the token is passed to each task that needs it

scripts/install.sh
7PH, line 1: I know it's bad, but we now have an install script for unix (cirrus) + windows (azure). Because gulp was dropping the npm token from the config before installing. I could verify this using an interactive terminal in Cirrus CI

7PH (approved):

PR #343, opened 2024-01-24
Revert MMF changes for safe patch release until minor release

7PH:
The goal here is to keep this PR open in case we have (really) unexpected issues after SQ release


7PH:
Closing this PR as there were no bug report on Community since the release, Wednesday

PR #351, opened 2024-02-29
VSTS-338 Bump SQ V4 task versions

7PH:
Follow-up on https://github.com/SonarSource/sonar-scanner-vsts/pull/346 which will let us release a new extension version fixing compatibility with V4 tasks


stanislavhh (approved):
LGTM

PR #353, opened 2024-03-06
VSTS-355 Ignore specified JDK 11 if server does not support it

7PH:
We can not just switch to Java 17 by default, because we would break users that do not have Java 17 installed in their agents.

Instead, we:
- Ignore the specified Java 11 if we know the server does not support it (SonarQube >= 10.4). In that case, switch to using the default JAVA_HOME.
- Add a warning message to the task when we ignore it and tell users to update the task to specify jdkversion

### DIFF
--- a/commonv5/ts/helpers/__tests__/java-version-resolver-test.ts
+++ b/commonv5/ts/helpers/__tests__/java-version-resolver-test.ts
@@ -1,61 +1,72 @@
 import * as tl from "azure-pipelines-task-lib/task";
+import { EndpointType } from "../../sonarqube/Endpoint";
+import { JdkVersionSource, TaskVariables } from "../constants";
 import JavaVersionResolver from "../java-version-resolver";
-import { TaskVariables } from "../constants";
+
+const MOCKED_JAVA_VARIABLES = {
+  [JdkVersionSource.JavaHome]: "/opt/bin/java/bin",
+  [JdkVersionSource.JavaHome11]: "/opt/bin/java11/bin",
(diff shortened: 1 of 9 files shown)


7PH:
Adding you as secondary reviewer @stanislavhh , in case you have time to take a look at this PR

commonv5/ts/helpers/__tests__/java-version-resolver-test.ts
```diff
@@ -1,61 +1,71 @@
...
-    jest.spyOn(tl, "getVariable").mockReturnValueOnce(undefined); // JAVA_HOME_11_X64
-
-    const actualJavaPath = JavaVersionResolver.lookupVariable(jdkSource);
-
-    expect(actualJavaPath).toBe(undefined);
-  });
+  it.each([
+    [EndpointType.SonarQube, undefined, JdkVersionSource.JavaHome11, "/opt/bin/java11/bin"],
+    [EndpointType.SonarQube, "9.9.0", JdkVersionSource.JavaHome11, "/opt/bin/java11/bin"],
+    [EndpointType.SonarQube, "9.9.0", JdkVersionSource.JavaHome17, "/opt/bin/java17/bin"],
```
lucas-paulger-sonarsource, line 37: out of curiosity, why do we not test SQ 10.4 here?

7PH: it's a good point, I'll add a test for SQ 10.4 + Java 17 to be more exhaustive

lucas-paulger-sonarsource (approved):
changes look good to me, validated in azure with SC test extension 🙌🏻

PR #355, opened 2024-03-15
VSTS-335 Drop PowerShell SQ tasks (V3)

7PH:
This PR drops V3 of SonarQube tasks, written in PowerShell. 

These tasks were deprecated [more than 6 years ago](https://github.com/SonarSource/sonar-scanner-vsts/commit/0fdc0266e8cce16f59896057de04e4c87305ea88). This PR drops all the code associated to these tasks.


config/paths.js
lucas-paulger-sonarsource, line 1: what caused the formatting change here? I see the only change was removing `old:` references so seems fine (but we can improve formatting in the CI to avoid this I think)

lucas-paulger-sonarsource (approved):
LGTM

lucas-paulger-sonarsource:
I think you can also drop`extensions/sonarqube/scanner-cli` as it is only used in the old build task

PR #357, opened 2024-03-26
VSTS-360 Add support to select JDK21 on hosted agents

7PH:
Add support to select JDK21 in analyze tasks of both the SonarCloud and SonarQube extensions. It is indeed now available in the hosted agents
image

through 
image
image


lucas-paulger-sonarsource (approved):
validated, I see the new option on SC, and the scan ran with the correct java version
`INFO: Java 21.0.2 Eclipse Adoptium (64-bit)`


Screenshot 2024-03-26 at 17 38 31

PR #359, opened 2024-04-04
change defaultvalue to reflect same changes in sonarcloud

ThisIsNotTheUserYouAreLookingFor:
Hi there, motives for change:
* current implementation defaults to java 11, which is not enough for the latest SonarQube version. See https://community.sonarsource.com/t/azure-pipelines-scanner-suddenly-requires-a-newer-java-version/112504

### DIFF
--- a/extensions/sonarqube/tasks/analyze/v5/task.json
+++ b/extensions/sonarqube/tasks/analyze/v5/task.json
@@ -20,7 +20,7 @@
       "name": "jdkversion",
       "type": "pickList",
       "label": "JDK version source for analysis",
-      "defaultValue": "JAVA_HOME_11_X64",
+      "defaultValue": "JAVA_HOME_17_X64",
       "required": true,
       "options": {
         "JAVA_HOME": "Use JAVA_HOME",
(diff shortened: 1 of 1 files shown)


7PH:
Hello and thank you for contributing to this repository @ThisIsNotTheUserYouAreLookingFor !

Unfortunately, we can not make the specific change you are requesting, because users on SQ < 10.5 that do not have Java 17 installed in their CI environment will be instantly broken by this update.

We are planning to release a new major version of the extension with version 6 of the SonarQube tasks and the scope of this task is to rethink the logic to which java version it defaults to (for sure, it will not default to Java 8). In the meantime, I would recommend not to rely on the default value, and specifying `jdkversion` in your pipeline should be considered a requirement. 

I am closing this PR as we can't merge this change for now, but please feel free to follow the planning for us implementing this task in our [internal ticket VSTS-356](https://sonarsource.atlassian.net/browse/VSTS-356)

sliekens:
@7PH  for my own understanding, why must the Java version on the build agent be aligned with the SQ server version?

7PH:
Hello @sliekens 

> why must the Java version on the build agent be aligned with the SQ server version?

When you are scanning a project, some `jars` are dynamically downloaded from your SonarQube instance to be run locally. These `jars` are shipped with SonarQube and are therefore tied to the SonarQube version you are using. 

Over time, we have been updating these `jars` and compiling them with increasing Java versions, following the industry. That means that as you update SonarQube, you need to be able to execute the `jars` it comes with, meaning having a JRE of at least the version we compiled it with.

If you are looking to dig more into this, I would recommend reading [the announcement we did to deprecate Java 11](https://community.sonarsource.com/t/java-11-is-deprecated-as-a-runtime-env-to-scan-your-projects/96597)

Hope my explanation helps

PR #365, opened 2024-05-24
VSTS-268 Bundle and tree-shake extension task code

7PH:
Using esbuild to inline all dependencies (and drop `node_modules` from task folders)

We need to to add a workaround [for shelljs because we can not bundle it](https://github.com/7PH/azure-devops-extension-bundler-shelljs-poc)

## Size reduction

Size before this PR, on master branch with V1+V4+V5
```text
19M     dist/sonar-scanner-vsts-1.46.0-sonarcloud.vsix
37M     dist/sonar-scanner-vsts-5.20.0-sonarqube.vsix
```

Size after this PR
```text
3.4M    dist/sonar-scanner-vsts-1.46.0-sonarcloud.vsix
6.8M    dist/sonar-scanner-vsts-5.20.0-sonarqube.vsix
```

(After rebase with the 2 new versions, V1+V2+V4+V5+V6):
```text
5.0M    ./dist/sonar-scanner-vsts-2.0.0-sonarcloud.vsix
9.1M    ./dist/sonar-scanner-vsts-6.0.0-sonarqube.vsix
```

## Build time

On my machine, before
```text
[12:24:35] Finished 'default' after 34 s
```

After
```text
[12:20:45] Finished 'default' after 12 s
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,5 +1,6 @@
 const path = require("path");
 const gulp = require("gulp");
+const gulpFile = require("gulp-file");
 const fs = require("fs-extra");
 const yargs = require("yargs");
 const gulpReplace = require("gulp-replace");
@@ -110,11 +111,11 @@ gulp.task("build:typescript", () => {
  * Bundle tasks
  */
 gulp.task("build:bundle", async () => {
-  const tasks = globby.sync(["src/extensions/*/tasks/*/v*/*.ts"]);
+  const tasks = globby.sync(path.join(SOURCE_DIR, "extensions", "*", "tasks", "*", "v*", "*.ts"));
   for (const task of tasks) {
     const [extension, , taskName, version] = task.split(path.sep).slice(-5);
     const commonFolder = getTaskCommonFolder(taskName, version);
-    const outFilePath = task.replace(/^src/, BUILD_DIR).replace(/\.ts$/, ".js");
+    const outFilePath = task.replace(SOURCE_DIR, BUILD_DIR).replace(/\.ts$/, ".js");
 
     // eslint-disable-next-line import/no-dynamic-require
     const esbuildConfig = require(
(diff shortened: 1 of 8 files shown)


gulpfile.js
```diff
@@ -151,28 +176,17 @@ gulp.task("build:copy", () => {
...
-      .pipe(gulp.dest(path.join(outPath, "node_modules")));
-    streams.push(copyNodeModulesStream);
+    /**
+     * Hotfix for shelljs dependency
+     * @see https://github.com/microsoft/azure-pipelines-task-lib/issues/942#issuecomment-1904939900
+     */
+    const createShellJsDummyFile = gulpFile("index.js", "", { src: true }).pipe(
+      gulp.dest(path.join(outPath, "node_modules", "shelljs")),
+    );
+    streams.push(createShellJsDummyFile);
```
7PH, line 189: shelljs doesn't support being bundled through `esbuild`, but is imported by the azure task lib. Without this fix, we can build but the task execution fails at runtime because the azure task lib always require `shelljs`, even though we are not using any exported function which uses this dependency. 

More explanation in this [POC](https://github.com/7PH/sonar-scanner-npm-mmf-3712-poc) I made to reproduce the behavior, issue and workaround

src/common/latest/esbuild.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  bundle: true,
+  platform: "node",
+  format: "cjs",
+  target: "node10",
```
lucas-paulger-sonarsource, line 5: can we not target node16 to align with the latest runner? or we can do that later in its own PR

gulpfile.js
lucas-paulger-sonarsource: we can also include the real shelljs into node_modules. How can we be sure we aren't using it at all on all systems?

I see it referenced inside the task.ts of the lib
https://github.com/microsoft/azure-pipelines-task-lib/blob/7c0de0dba61afb899068c35d88257017bd3b4736/node/task.ts#L2C8-L2C13

If you're confident it isn't used for any platform, I'm okay with stubbing it like this.

lucas-paulger-sonarsource (approved):

src/common/latest/esbuild.config.js
7PH: Definitively true, I will update it!

gulpfile.js
7PH: It's not used, if we were using at some point, yes it would fail in QA and production (we should spot it in QA),

I considered adding shelljs but it's 1.7MB in each shipped task, so figured, since it's a workaround anyway we can save 2M * 15 tasks by adding this 0B file.

src/common/latest/esbuild.config.js
7PH: TY, didn't think about it, [added it to the other ticket](https://sonarsource.atlassian.net/browse/VSTS-376) for bumping the agent versions (because ATM this is still running on Node 10)

PR #367, opened 2024-05-27
NO-JIRA [Do not merge] Fix build

7PH:


src/extensions/sonarcloud/vss-extension.json
```diff
@@ -2,7 +2,7 @@
   "manifestVersion": 1,
   "id": "sonarcloud",
   "name": "SonarCloud",
-  "version": "1.46.0",
+  "version": "5.50.11",
```
lucas-paulger-sonarsource, line 5: afaik, we can bump the task version without needing to bump extension version and it should still bust the task cache properly.

I wanted to suggest that we could implement bumping the patch version per run. and when we release we bump minor and reset the patch back to 0

lucas-paulger-sonarsource: my guess is that the assumption was by deleting and re-installing the extension for ITs , it was clearing the cache.

lucas-paulger-sonarsource: you would think the extension cache would supersede the task cache but it seems that's not the case

7PH:
Issue fixed itself. Closing this PR unless the issue happens again, but I'd monitor QA in the next runs to be sure it doesn't happen again.

PR #369, opened 2024-05-27
NO-JIRA Fix build

7PH:
Round 2, since there is actually an issue with the build ATM


azp/helpers.ps1
7PH, line 1: The azure pipeline runs on ubuntu, so there's no need for having these helper scripts as powershell, they are now gulp tasks:
- `ci:azure:hotfix-extensions-version`
- `ci:azure:hotfix-extensions-version`

gulpfile.js
```diff
@@ -361,6 +361,96 @@ gulp.task("sonarqube:scan", async () => {
...
+});
+
+gulp.task("ci:azure:hotfix-tasks-version", () => {
+  const buildNumber = process.env.BUILD_BUILDID;
+
+  if (!buildNumber) {
+    throw new Error("Missing build number");
+  }
+
+  const tasks = globby.sync(["src/extensions/*/tasks/*/v*/*.json"]);
```
lucas-paulger-sonarsource, line 400: yay updating the task version as well is nice!

gulpfile.js
```diff
@@ -361,6 +361,96 @@ gulp.task("sonarqube:scan", async () => {
...
+  return mergeStream(
+    tasks.map((task) =>
+      gulp
+        .src(task)
+        .pipe(
+          gulpJsonEditor((json) => ({
+            ...json,
+            version: {
+              ...json.version,
+              Patch: parseInt(buildNumber, 10),
```
lucas-paulger-sonarsource, line 411: I'm not familiar with `Patch` in the version object. I assume it appends the Patch version to equal the build number

gulpfile.js
```diff
@@ -361,6 +361,96 @@ gulp.task("sonarqube:scan", async () => {
...
+      "build",
+      "extensions",
+      extension,
+      "vss-extension.json",
+    );
+    console.log(`Fetched ext version ${version} for ${extension}`);
+    console.log(`##vso[task.setvariable variable=${extensionPrefix}_EXT_NAME]${extensionManifest}`);
+    console.log(
+      `##vso[task.setvariable variable=${extensionPrefix}_VERSION;isOutput=true]${version}`,
+    );
```
lucas-paulger-sonarsource, line 446: this log syntax mirrors what we would see from the Azure task. Does it need to match or just for astetics? 

if not, I would say to keep it distinctly different. But if so I would want to know why it's needed :D

src/common/latest/analyze-task.ts
```diff
@@ -12,9 +11,11 @@ import { EndpointType } from "./sonarqube/Endpoint";
 import Scanner, { ScannerMode } from "./sonarqube/Scanner";
 
 export const analyzeTask: TaskJob = async (endpointType: EndpointType) => {
-  const rootPath = path.join(__dirname, "..");
+  const rootPath = __dirname;
```
lucas-paulger-sonarsource, line 14: we're back!

src/common/latest/analyze-task.ts
```diff
@@ -12,9 +11,11 @@ import { EndpointType } from "./sonarqube/Endpoint";
...
   const jdkVersionSource = tl.getInput("jdkversion", true) as JdkVersionSource;
 
+  tl.warning("THIS SHOULD BE SHOWN");
```
lucas-paulger-sonarsource, line 17: purpose?

src/common/sonarcloud-v1/analyze-task.ts
```diff
@@ -12,9 +11,12 @@ import { EndpointType } from "./sonarqube/Endpoint";
 import Scanner, { ScannerMode } from "./sonarqube/Scanner";
 
 export const analyzeTask: TaskJob = async (endpointType: EndpointType) => {
-  const rootPath = path.join(__dirname, "..");
+  const rootPath = __dirname;
   const jdkVersionSource = tl.getInput("jdkversion", true) as JdkVersionSource;
 
+  tl.debug(`CWD(): ${process.cwd()}`);
+  tl.debug(`rootPath: ${rootPath}`);
+
```
lucas-paulger-sonarsource, line 19: probably can drop these, unless you see value in keeping them

src/common/sonarcloud-v1/prepare-task.ts
```diff
@@ -17,12 +16,14 @@ import TaskReport from "./sonarqube/TaskReport";
...
 export const prepareTask: TaskJob = async (endpointType: EndpointType) => {
   const endpoint = Endpoint.getEndpoint(tl.getInput(endpointType, true), endpointType);
-  const rootPath = path.join(__dirname, "..");
+  const rootPath = __dirname;
 
   const scannerMode: ScannerMode = ScannerMode[tl.getInput("scannerMode")];
   const scanner = Scanner.getPrepareScanner(rootPath, scannerMode);
   const serverVersion = await getServerVersion(endpoint);
 
+  tl.debug(`>>> CWD(): ${process.cwd()}`);
```
lucas-paulger-sonarsource, line 25: same, cleanup if not needed

src/common/sonarqube-v4/__tests__/analyze-task-test.ts
```diff
@@ -48,7 +48,7 @@ it("should run scanner", async () => {
   await analyze.analyzeTask(EndpointType.SonarQube);
 
   expect(Scanner.getAnalyzeScanner).toHaveBeenCalledWith(
-    path.join(__dirname, "..", ".."),
+    path.join(__dirname, ".."),
```
lucas-paulger-sonarsource, line 51: shouldn't this be `__dirname`? seeing below
https://github.com/SonarSource/sonar-scanner-vsts/pull/369/files#diff-b61d2ef3250bb55d7f9192dfed56523dcb5e2027ee2e6fcd8f7ef004853eee06R27

we pass the rootpath, am I missing something?

gulpfile.js
7PH: Yep, it replaces [this `Patch`](https://github.com/SonarSource/sonar-scanner-vsts/blob/master/src/extensions/sonarqube/tasks/SonarQubePrepare/v5/task.json#L13) with the build number

gulpfile.js
7PH: This syntax is not aesthetic at all but it's needed because it's [the way to define variables in AzDO](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/set-variables-scripts?view=azure-devops&tabs=bash)

It's used [here](https://github.com/SonarSource/sonar-scanner-vsts/blob/master/azure-pipelines.yml#L159)

src/common/latest/analyze-task.ts
7PH: no purpose, it should be dropped, thanks for spotting

src/common/sonarqube-v4/__tests__/analyze-task-test.ts
7PH: We shouldn't because we need to go from `__tests__` folder to parent folder where `__dirname` is executed in `analyze-task.ts`

PR #373, opened 2024-05-29
NO-JIRA Drop v1/v2/v4/v5 repetition in favor of iterators

7PH:
Minor PR to clean up the references to each v1/v2/v4/v5 common folders and iterate instead


lucas-paulger-sonarsource (approved):
much cleaner 🙌🏻

PR #380, opened 2024-06-02
VSTS-377 Use latest version of NPM Scanner & Fix Next analysis project version

7PH:
- The project version on Next is invalid and uses the one from the root package.json, we should go and fetch the version from each `vss-extension.json`
- Uses the new Scanner for NPM v4


lucas-paulger-sonarsource (approved):

7PH:
Putting this in draft until `sonar-scanner-npm` is fixed

PR #385, opened 2024-06-18
SONARAZDO-386 Bump sonar-scanner-cli to v6

7PH:
[ticket](https://sonarsource.atlassian.net/browse/SONARAZDO-386)
- Bump the default scanner version to v6 for SQV2 and SCV6 tasks. 
- This will help mitigate the `ThrowableProxy` error due to `jGit` for SonarCloud users.
- This is a risky release (in rare cases, `sonar-scanner-cli v5 -> v6` is a breaking change (eg when the architecture can't be detected), but we released the new task versions soon, so failing now is acceptable IMO, PMs are OK with the risk as well.

Change validated on SQV6, SCV2, SQV5, SCV1

### DIFF
--- a/src/extensions/sonarcloud/tasks/SonarCloudAnalyze/v2/task.json
+++ b/src/extensions/sonarcloud/tasks/SonarCloudAnalyze/v2/task.json
@@ -9,7 +9,7 @@
   "author": "sonarsource",
   "version": {
     "Major": 2,
-    "Minor": 0,
+    "Minor": 1,
     "Patch": 0
   },
   "minimumAgentVersion": "3.218.0",
--- a/src/common/latest/config.ts
+++ b/src/common/latest/config.ts
@@ -1,6 +1,6 @@
 // When the user does not specify a specific version, these willl be the default versions used.
 const msBuildVersion = "6.2.0.85879";
-const cliVersion = "5.0.1.3006";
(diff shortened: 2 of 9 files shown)


lucas-paulger-sonarsource (approved):
LGTM, we will need to review the release pipeline to ensure it's still functional

src/extensions/sonarcloud/tasks/SonarCloudAnalyze/v2/task.json
```diff
@@ -10,7 +10,7 @@
   "version": {
     "Major": 2,
     "Minor": 0,
-    "Patch": 0
+    "Patch": 1
```
lucas-paulger-sonarsource, line 13: when we bump default scanner versions, is it patch or minor?

7PH: Yeah, I wasn't sure about this, IMO we're not adding any feature from the user POV so it's a patch? But I don't have a strong opinion on this

lucas-paulger-sonarsource: I was thinking maybe because the scanner-cli version is a major bump it might warrant a bit more "caution". I'm fine either way

PR #388, opened 2024-06-27
SONARAZDO-385 Allow offline execution for SQV6, SCV2 tasks

7PH:
- Embed default scanners in SQV6 and SCV2 tasks
  - This PR alsoimplements [SONARAZDO-351](https://sonarsource.atlassian.net/browse/SONARAZDO-351) to build a mocking architecture allowing to mock AzDO variables/input directly rather than assuming the invocation order
- Change build process to support independent msbuild vs scanner cli versions
- Drop `777` permission (replaced with `500`) on scanner files (this is failing SQ scan)

[SONARAZDO-351]: https://sonarsource.atlassian.net/browse/SONARAZDO-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/common/latest/__tests__/prepare-task-test.ts
+++ b/src/common/latest/__tests__/prepare-task-test.ts
@@ -3,16 +3,30 @@ import * as toolLib from "azure-pipelines-tool-lib/tool";
 import { Guid } from "guid-typescript";
 import * as path from "path";
 import { SemVer } from "semver";
-import { scanner } from "../config";
-import { JdkVersionSource, TaskVariables } from "../helpers/constants";
+import { JdkVersionSource } from "../helpers/constants";
 import * as request from "../helpers/request";
+import { AzureTaskLibMock } from "../mocks/AzureTaskLibMock";
+import { AzureToolLibMock } from "../mocks/AzureToolLibMock";
 import * as prept from "../prepare-task";
 import Endpoint, { EndpointType } from "../sonarqube/Endpoint";
 import Scanner, { ScannerCLI, ScannerMSBuild, ScannerMode } from "../sonarqube/Scanner";
 import TaskReport from "../sonarqube/TaskReport";
 
+jest.mock("azure-pipelines-task-lib/task");
+jest.mock("azure-pipelines-tool-lib/tool");
+
+jest.mock("fs-extra", () => ({
+  ...jest.requireActual("fs-extra"),
+  chmod: jest.fn(),
+}));
+
+const azureTaskLibMock = new AzureTaskLibMock();
+const azureToolLibMock = new AzureToolLibMock();
+
 beforeEach(() => {
   jest.restoreAllMocks();
+  azureTaskLibMock.reset();
+  azureToolLibMock.reset();
 });
 
 describe("branchFeatureSupported", () => {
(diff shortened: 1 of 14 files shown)


src/common/latest/__tests__/prepare-task-test.ts
```diff
@@ -3,16 +3,30 @@ import * as toolLib from "azure-pipelines-tool-lib/tool";
...
 import * as request from "../helpers/request";
+import { AzureTaskLibMock } from "../mocks/AzureTaskLibMock";
+import { AzureToolLibMock } from "../mocks/AzureToolLibMock";
 import * as prept from "../prepare-task";
 import Endpoint, { EndpointType } from "../sonarqube/Endpoint";
 import Scanner, { ScannerCLI, ScannerMSBuild, ScannerMode } from "../sonarqube/Scanner";
 import TaskReport from "../sonarqube/TaskReport";
 
+jest.mock("azure-pipelines-task-lib/task");
+jest.mock("azure-pipelines-tool-lib/tool");
```
7PH, line 16: For some reason, we need to use `jest.mock()` on these libs both in the mock helper and in the utility functions that use them. I'm happy to optimize this if you have any idea about how to, because if we're using the mock helpers, then for sure we want these libs mocked!

lucas-paulger-sonarsource: I think that's fine 👍🏻 we were overriding the method responses below and this will basically catch if we need to mock something specific

src/common/latest/__tests__/prepare-task-test.ts
```diff
@@ -103,12 +109,14 @@ describe("downloading the scanner", () => {
...
-      jest.spyOn(tl, "getInput").mockImplementationOnce(() => "mock-organization");
-      jest.spyOn(tl, "getInput").mockImplementationOnce(() => msBuildVersion);
-      jest.spyOn(tl, "getInput").mockImplementationOnce(() => cliVersion);
-      jest.spyOn(tl, "getInput").mockImplementationOnce(() => scannerMode);
+      azureTaskLibMock.setInputs({
+        SonarQube: "SQ",
+        jdkversion: JdkVersionSource.JavaHome,
+        organization: "mock-organization",
+        msBuildVersion,
+        cliVersion,
```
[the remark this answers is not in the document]
lucas-paulger-sonarsource: My only concern will be maintaining these mock libraries, but I agree better not to need to care about the import order. Lets go with it 🙌🏻

lucas-paulger-sonarsource (approved):
LGTM 🚀

src/common/latest/config.ts
```diff
@@ -1,6 +1,6 @@
 // When the user does not specify a specific version, these willl be the default versions used.
 const msBuildVersion = "6.2.0.85879";
-const cliVersion = "6.0.0.4432";
+const cliVersion = "6.1.0.4477";
```
lucas-paulger-sonarsource, line 3: do these versions still align?

7PH: They don't have anymore thanks to [this commit](https://github.com/SonarSource/sonar-scanner-azdo/pull/388/commits/a3c02628133ad6ce2d2418b20fc786416193e228) 😎

PR #389, opened 2024-06-28
SONARAZDO-387 Drop the scripts/install.sh script

7PH:
This fixes debt added when we had to solve [anonymous access to repox](https://github.com/SonarSource/sonar-scanner-azdo/pull/339).

The Repox token isn't forwarded to npm scripts when run in Cirrus CI. I could not find any documentation explaining why. This PR aligns how we handle NPM configuration with the [SonarScanner for NPM project](https://github.com/SonarSource/sonar-scanner-npm) by using `.npmrc` to hold the npm configuration that should be used in Cirrus.


7PH:
Not satisfied with the solution, closing for now.

PR #397, opened 2024-07-31
SONARAZDO-394 Upgrade node-fetch to v3

7PH:
This comes from a dogfood thread / support request

We can finally upgrade to `node-fetch@3` on the Node16 tasks

[ticket](https://sonarsource.atlassian.net/browse/SONARAZDO-394)


lucas-paulger-sonarsource (approved):
LGTM

PR #398, opened 2024-08-02
BUILD-5506 Fix artifactsToDownload value

7PH:
[ticket](https://sonarsource.atlassian.net/browse/BUILD-5506)


7PH:
The release process now works

PR #404, opened 2024-09-17
SONARAZDO-405 Replace node-fetch with axios

7PH:
Please refer to [the ticket](https://sonarsource.atlassian.net/browse/SONARAZDO-405) for context.

### Intro

This PR drops `node-fetch` and uses `axios`, used in SonarQube and Scanner for NPM.

### Validation

<details>

SQ, LTS
image

SQ, latest
image

SC (proxy)
image

SC
image
</details>


src/common/latest/sonarqube/Endpoint.ts
```diff
@@ -75,8 +71,14 @@ export default class Endpoint {
 
     // When proxy is used we use HttpsProxyAgent or HttpProxyAgent to handle it.
     if (proxyUrl) {
-      const AgentClass = endpointUrl.startsWith("https://") ? HttpsProxyAgent : HttpProxyAgent;
-      options.agent = new AgentClass({ proxy: proxyUrl });
+      // We need to set proxy to false to avoid conflicts with agent proxy configuration
+      options.proxy = false;
```
7PH, line 75: refer to https://github.com/coder/vscode-coder/commit/4dc0db909446feaa49490b7b580a1f12461a06a1

src/common/latest/sonarqube/Endpoint.ts
```diff
@@ -31,41 +30,38 @@ export default class Endpoint {
...
-    const options: Partial<RequestInit> = {
-      method: "get",
-      signal,
-    };
-
-    // Add HTTP auth from this.auth
-    options.headers = {
-      Authorization:
-        "Basic " + Buffer.from(`${this.auth.username}:${this.auth.password}`).toString("base64"),
+  toAxiosOptions(): AxiosRequestConfig {
```
7PH, line 50: Passing the endpoint URL doesn't really makes sense as we can directly access `this.url`

src/common/latest/helpers/__tests__/request-test.ts
```diff
@@ -1,82 +1,172 @@
...
-    expect(tl.debug).toHaveBeenCalledWith(
-      `[SQ] API GET '/api/some-api' failed, error is Request timeout`,
-    );
+      const response = await get(endpoint, "/api/server/version");
+      expect(response).toEqual("1.2.3.4");
+
+      const args = axiosMock.history.get[0];
+      const { httpAgent, httpsAgent } = args;
+      expect(httpAgent).toBeUndefined();
+      expect(httpsAgent).toBeInstanceOf(HttpsProxyAgent);
```
stanislavhh, line 143: I think it won't hurt to add:
```     
 expect(httpsAgent.proxy.origin).toBe(process.env.HTTP_PROXY);
```

src/common/latest/sonarqube/Endpoint.ts
stanislavhh: the only difference is that we were passing "fullUrl" that included path + query, however, I don't think we need it

stanislavhh (approved):
Thank you for the detailed validation screenshots!
Code looks good to me ✔️  only 1 question.

What is the plan for this PR? Is the plan to include this fix to latest public version or introduce it in scope of major? 

I don't think I'll be efficient in validation, and since you provided SS, I guess I'll skip them. Let me know if you think otherwise/I can help you with some scenarios.

7PH:
Hey @stanislavhh , thanks for the review!

The plan is to include this in the latest public version on master (unless S4NET v9 is released this week, which is unlikely from the latest update I had from them)

I'll take time to triple check your comments, make another validation round and merge!

src/common/latest/sonarqube/Endpoint.ts
```diff
@@ -31,41 +30,38 @@ export default class Endpoint {
...
   }
 
-  public get auth(): { username: string; password: string } {
+  public get auth(): { username: string; password?: string } {
     // If using user/password
     if (!this.data.token && this.data.password && this.data.password.length > 0) {
       return { username: this.data.username, password: this.data.password };
     }
     // Using token
-    return { username: this.data.token || this.data.username, password: "" };
```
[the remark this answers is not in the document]
7PH: I reverted to what we used before, and it works because it's been validated with token only (which sets password as undefined).

But when checking the types, it looks like we actually should not omit `password`. I'll double check this (why isn't the type check finding this, and what should be used)

PR #405, opened 2024-09-18
SONARAZDO-406 Fix confusing log messages & Introduce INFO logs

7PH:
[ticket](https://sonarsource.atlassian.net/browse/SONARAZDO-406)

- There is little functional impact (some log type are changed from debug to info, or from/to warnings/errors). Functional changes are highlighted below.
- Due to the logging changes, the initial coverage was very low. I added a few unit tests to improve the coverage for this PR.

### DIFF
--- a/src/common/latest/helpers/api.ts
+++ b/src/common/latest/helpers/api.ts
@@ -1,4 +1,3 @@
-import * as tl from "azure-pipelines-task-lib/task";
 import Endpoint from "../sonarqube/Endpoint";
 import {
   Measure,
@@ -7,7 +6,8 @@ import {
   MetricsResponse,
   ProjectStatus,
 } from "../sonarqube/types";
-import { RequestData, get } from "./request";
+import { log, LogLevel } from "./logging";
+import { get, RequestData } from "./request";
 import { waitFor } from "./utils";
 
 export const RETRY_MAX_ATTEMPTS = 3;
@@ -24,18 +24,18 @@ export async function fetchWithRetry<T>(
       return await get<T>(endpoint, path, query);
     } catch (error) {
       attempts++;
-      tl.debug(`[SQ] API GET '${path}' failed (attempt ${attempts}/${RETRY_MAX_ATTEMPTS})`);
+      log(LogLevel.DEBUG, `API GET '${path}' failed (attempt ${attempts}/${RETRY_MAX_ATTEMPTS})`);
       await waitFor(RETRY_DELAY);
     }
   }
-  throw new Error(`[SQ] API GET '${path}' failed, max attempts reached`);
+  throw new Error(`API GET '${path}' failed, max attempts reached`);
 }
 
 export async function fetchProjectStatus(
   endpoint: Endpoint,
   analysisId: string,
 ): Promise<ProjectStatus> {
-  tl.debug(`Retrieve Analysis id '${analysisId}.'`);
+  log(LogLevel.DEBUG, `Retrieve Analysis id '${analysisId}.'`);
 
   try {
     const { projectStatus } = await get<{ projectStatus: ProjectStatus }>(
@@ -48,9 +48,9 @@ export async function fetchProjectStatus(
     return projectStatus;
   } catch (error) {
     if (error?.message) {
-      tl.error(`Error retrieving analysis: ${error.message}`);
+      log(LogLevel.ERROR, `Error retrieving analysis: ${error.message}`);
     } else if (error) {
-      tl.error(`Error retrieving analysis: ${JSON.stringify(error)}`);
+      log(LogLevel.ERROR, `Error retrieving analysis: ${JSON.stringify(error)}`);
     }
     throw new Error(`Could not fetch analysis for ID '${analysisId}'`);
   }
@@ -78,9 +78,9 @@ export async function fetchMetrics(
     );
   } catch (error) {
     if (error?.message) {
-      tl.error(error.message);
+      log(LogLevel.ERROR, error.message);
     } else if (error) {
-      tl.error(JSON.stringify(error));
+      log(LogLevel.ERROR, JSON.stringify(error));
     }
 
     throw new Error(`Could not fetch metrics`);
@@ -95,10 +95,11 @@ export async function fetchComponentMeasures(
     const response = await get<MeasureResponse>(endpoint, "/api/measures/component", data);
     return response.component.measures;
   } catch (error) {
-    if (error?.message) {
-      tl.debug("Error fetching component measures: " + error.message);
-    } else if (error) {
-      tl.debug("Error fetching component measures: " + JSON.stringify(error));
+    if (error) {
+      log(
+        LogLevel.INFO,
+        "Error fetching component measures: " + (error.message ?? JSON.stringify(error)),
+      );
     }
 
     throw new Error(`Could not fetch component measures`);
--- a/src/common/latest/helpers/logging.ts
+++ b/src/common/latest/helpers/logging.ts
@@ -1,0 +1,56 @@
+/*
+ * sonar-scanner-npm
+ * Copyright (C) 2022-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import * as tl from "azure-pipelines-task-lib/task";
+import { EndpointType } from "../sonarqube";
+
+export enum LogLevel {
+  DEBUG = "DEBUG",
+  INFO = "INFO",
+  WARN = "WARN",
+  ERROR = "ERROR",
+}
+
+const LOG_MESSAGE_PADDING = 7;
+
+let endpointType: EndpointType | undefined;
+
+export function log(level: LogLevel, message: string) {
+  const levelStr = `[${level}]`.padEnd(LOG_MESSAGE_PADDING);
+  const msgStr = [levelStr, `${endpointType ?? EndpointType.SonarQube}:`, message].join(" ");
+  switch (level) {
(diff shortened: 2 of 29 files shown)


src/common/latest/helpers/api.ts
```diff
@@ -97,10 +97,11 @@ export async function fetchComponentMeasures(
...
   } catch (error) {
-    if (error?.message) {
-      tl.debug("Error fetching component measures: " + error.message);
-    } else if (error) {
-      tl.debug("Error fetching component measures: " + JSON.stringify(error));
+    if (error) {
+      log(
+        LogLevel.INFO,
+        "Error fetching component measures: " + (error.message ?? JSON.stringify(error)),
+      );
```
7PH, line 104: Changed from `DEBUG` to `INFO`

src/common/latest/helpers/request.ts
```diff
@@ -34,18 +35,18 @@ export async function get<T>(
...
+    throw new Error(`API GET '${path}' failed, error is ${error.message}`);
   } finally {
     clearTimeout(timeout);
   }
 }
 
 export async function getServerVersion(endpoint: Endpoint): Promise<semver.SemVer> {
   const serverVersion = await get<string>(endpoint, "/api/server/version", false);
-  tl.debug(`[SQ] Server version: ${serverVersion}`);
+  log(LogLevel.INFO, `Server version: ${serverVersion}`);
```
7PH, line 50: Changed from `DEBUG` to `INFO`

src/common/latest/prepare-task.ts
```diff
@@ -141,7 +143,8 @@ async function isDefaultBranch() {
       AzureProvider.Bitbucket,
     ].includes(provider)
   ) {
-    tl.debug(
+    log(
+      LogLevel.INFO,
```
7PH, line 147: Changed from `DEBUG` to `INFO`

src/common/latest/prepare-task.ts
```diff
@@ -168,10 +171,10 @@ export async function getDefaultBranch(collectionUrl: string): Promise<string |
       tl.getVariable(AzureBuildVariables.BuildRepositoryName),
       tl.getVariable("System.TeamProject"),
     );
-    tl.debug(`Default branch of this repository is '${repo.defaultBranch}'`);
+    log(LogLevel.INFO, `Default branch of this repository is '${repo.defaultBranch}'`);
     return repo.defaultBranch;
   } catch (e) {
-    tl.debug("Unable to get default branch, defaulting to 'master': " + e);
+    log(LogLevel.INFO, "Unable to get default branch, defaulting to 'master': " + e);
     return DEFAULT_BRANCH_REF;
```
7PH, line 178: Changed from `DEBUG` to `INFO`

src/common/latest/publish-task.ts
```diff
@@ -48,12 +49,12 @@ export const publishTask: TaskJob = async (_endpointType: EndpointType) => {
...
   if (!taskReports.length) {
-    tl.warning("No analyses found in this build! Please check your build configuration.");
+    log(LogLevel.WARN, "No analyses found in this build! Please check your build configuration.");
   } else {
-    tl.debug(`Number of analyses in this build: ${taskReports.length}`);
+    log(LogLevel.DEBUG, `Number of analyses in this build: ${taskReports.length}`);
   }
 
-  tl.debug(`Overall Quality Gate status: ${globalQualityGateStatus}`);
+  log(LogLevel.INFO, `Overall Quality Gate status: ${globalQualityGateStatus}`);
```
7PH, line 57: Changed from `DEBUG` to `INFO`

src/common/latest/publish-task.ts
```diff
@@ -111,7 +113,11 @@ export async function getReportForTask(
   timeoutSec: number,
 ): Promise<string> {
   try {
-    const task = await Task.waitForTaskCompletion(endpoint, taskReport.ceTaskId, timeoutSec, 1000);
+    const task = await waitForTaskCompletion(endpoint, taskReport.ceTaskId, timeoutSec, 1000);
+    for (const warning of task.warnings ?? []) {
+      log(LogLevel.INFO, `Analysis succeeded with warning: ${warning}`);
+    }
```
7PH, line 119: Add analysis warning logs as `INFO` logs, for each warning

src/common/latest/run.ts
```diff
@@ -1,13 +1,16 @@
...
+export async function runTask(fun: TaskJob, taskName: string, endpointType: EndpointType) {
+  setEndpointType(endpointType);
 
-export async function runTask(fun: TaskJob, taskName: string, endpoint: EndpointType) {
   try {
-    await fun(endpoint);
+    await fun(endpointType);
   } catch (err) {
-    tl.warning(`Error while executing ${endpoint}:${taskName} task: ${err.message}`);
+    log(LogLevel.ERROR, `Error while executing task ${taskName}: ${err.message}`);
```
7PH, line 13: Changed from `WARN` to `ERROR` (because we fail the task after this error)

src/common/latest/sonarqube/Endpoint.ts
```diff
@@ -67,14 +68,10 @@ export default class Endpoint {
     const proxyUrl = getProxyFromURI(new URL(endpointUrl))
       ? getProxyFromURI(new URL(endpointUrl))
       : undefined;
-    if (proxyUrl) {
-      tl.debug("Using proxy agent from environment: " + proxyUrl);
-    } else {
-      tl.debug("Not using a proxy agent");
-    }
```
7PH, line 74: Dropped and moved to line `74`

src/common/latest/sonarqube/TaskReport.ts
```diff
@@ -70,7 +71,8 @@ export default class TaskReport {
     let taskReportGlobResult: string[];
 
     if (endpoint.type === EndpointType.SonarQube && semver.satisfies(serverVersion, "<7.2.0")) {
-      tl.debug(
+      log(
+        LogLevel.INFO,
         "SonarQube version < 7.2.0 detected, falling back to default location(s) for report-task.txt file.",
```
7PH, line 76: Changed from `DEBUG` to `INFO`

src/common/latest/helpers/logging.ts
```diff
@@ -0,0 +1,56 @@
...
+  const msgStr = [levelStr, `${endpointType ?? EndpointType.SonarQube}:`, message].join(" ");
+  switch (level) {
+    case LogLevel.ERROR:
+      tl.error(msgStr);
+      break;
+    case LogLevel.WARN:
+      tl.warning(msgStr);
+      break;
+    case LogLevel.INFO:
+      console.log(msgStr);
```
lucas-paulger-sonarsource, line 46: looking at the validate-ci task where the tests are run, I am seeing duplicate for each console.log

https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=100618&view=logs&j=f1338668-1e5b-508a-0652-45679475677e&t=41df0c21-6a86-5c69-c5a6-7f399df43a84&l=518

lucas-paulger-sonarsource:
Don't forget to rebase, I still see V4 generated for SQ

src/common/latest/helpers/request.ts
lucas-paulger-sonarsource: I think I still see some references to `[SQ]` prefix,  we could clean it up across the board
https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=100618&view=logs&j=f1338668-1e5b-508a-0652-45679475677e&t=41df0c21-6a86-5c69-c5a6-7f399df43a84&l=1742

src/common/latest/sonarqube/Endpoint.ts
lucas-paulger-sonarsource: I guess there's no need to log when the proxy agent isn't used? 

only reason I could see is calling out to a user when they believe PROXY is used, but indeed we skip it

src/common/latest/helpers/request.ts
lucas-paulger-sonarsource: oh ignore me, I see that the test runs SC v1 and we only updated latest, which makes sense

src/common/latest/run.ts
```diff
@@ -1,13 +1,16 @@
...
-export type TaskJob = (endpoint: EndpointType) => Promise<void>;
+export type TaskJob = (endpointType: EndpointType) => Promise<void>;
+
+export async function runTask(fun: TaskJob, taskName: string, endpointType: EndpointType) {
+  setEndpointType(endpointType);
 
-export async function runTask(fun: TaskJob, taskName: string, endpoint: EndpointType) {
   try {
-    await fun(endpoint);
+    await fun(endpointType);
```
lucas-paulger-sonarsource, line 11: how does the user set the log level override for DEBUG (verbose)

I would expect something like:
https://github.com/SonarSource/sonar-scanner-npm/blob/master/src/scan.ts#L50C28-L50C57

I see that by default the user sees INFO/WARN/ERROR

lucas-paulger-sonarsource (changes requested):
Overall great improvement! the logs look much cleaner (I did FV) but I think you missed having a way for the user to set the logLevel (e.g. debug/verbose) Unless I am missing it 🤔

src/common/latest/helpers/logging.ts
7PH: I don't think they're duplicate, they're different (similar) tests 🤔

src/common/latest/run.ts
7PH: They pass it through Azure DevOps when they check this
image

I didn't find a way to set the pipeline in debug mode from within a task  🤔so the user can choose 2 log levels:
- debug, he'll see everything
- default, he'll see info, warn and errors

That's why I also dropped the TRACE logs (compared to the NPM implem) WDYT?

src/common/latest/sonarqube/Endpoint.ts
7PH: My thinking was, we can still know because of the presence/absence of this debug log? Also found it annoying to have this log on every request, but idk, wdyt?

lucas-paulger-sonarsource: that makes sense, reduce the noise, and if the user sets the proxy they should see something about it, if not it's not configured properly 👍🏻

src/common/latest/helpers/logging.ts
lucas-paulger-sonarsource: yeah, I think you're right. Usually I would prefer no logs at all in tests. If we care to know they're called, we can mock and assert it was called with the expected message. I think you did this for debug and error

```
    jest.mocked(tl.error).mockImplementation(() => {});
    jest.mocked(tl.debug).mockImplementation(() => {});
```

maybe we do the same for console.log since it's now the INFO level logging? WDYT?

src/common/latest/run.ts
lucas-paulger-sonarsource: ah great! I apparently always missed this flag 😅

lucas-paulger-sonarsource (approved):
Thanks for the callout on System Diagnostics. No blockers, only consideration is removing the console logs from tests

PR #408, opened 2024-09-19
SONARAZDO-397 Track extension size in pull requests

7PH:
If a pull request bumps one of the `.vsix` file size to >30MB, the `check_size` task will fail with an error:
image

BTW I think the limit should be 20MB (because that's the limit for TFS Server 2018), but we currently are over this limit already.


lucas-paulger-sonarsource (approved):
LGTM

PR #409, opened 2024-09-20
SONARAZDO-407 Allow more than one pull request to be open and run QA at the same time

7PH:
Combined with this setting in AzDO, this change should allow having multiple QA running simultaneously (with 1 max concurrent run)

image


lucas-paulger-sonarsource (approved):
LGTM

PR #415, opened 2024-09-26
SONARAZDO-398 Ensure we test code all platforms (Windows + UNIX) for .NET

7PH:
### Changes
- New pipeline and project keys mention what the E2E pipeline tests (SonarCloud, task version, language)
- Project names are normalized on SonarCloud 

### Additions
- Added a test for the old major task versions (SC V2) using CLI scanner provisioning
- Added a test for NET Core using Unix (we were only testing NET Framework on Windows)

### Impact
- The QA will run slower because of the lack of concurrency in our task runners (+4min)
- We now cover CLI provisioning, SC V2 tasks, .NET Core on Unix

### Screenshots 
Old E2E pipelines
image

New E2E pipelines
image

New E2E normalized project names
image

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -182,22 +182,6 @@ stages:
               AZDO_BASE_URL: $(AzdoBaseUrl)
               ITS_ORGA: $(AzdoItsOrga)
               ITS_PROJECT_NAME: $(AzdoItsProjectName)
-              ITS_DOTNET_PIPELINE_NAME: $(AzdoItsDotnetPipelineName)
-              ITS_DOTNET_SC_PROJECT_KEY: $(SonarCloudItsDotnetProjectKey)
-              ITS_DOTNET_EXPECTED_COVERAGE: $(DotnetExpectedCoverage)
-              ITS_DOTNET_EXPECTED_NCLOCS: $(DotnetExpectedNcLocs)
-              ITS_MAVEN_PIPELINE_NAME: $(AzdoItsMavenPipelineName)
-              ITS_MAVEN_SC_PROJECT_KEY: $(SonarCloudItsMavenProjectKey)
-              ITS_MAVEN_EXPECTED_COVERAGE: $(MavenExpectedCoverage)
-              ITS_MAVEN_EXPECTED_NCLOCS: $(MavenExpectedNcLocs)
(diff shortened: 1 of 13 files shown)


azure-pipelines.yml
```diff
@@ -182,22 +182,6 @@ stages:
...
-              ITS_MAVEN_EXPECTED_COVERAGE: $(MavenExpectedCoverage)
-              ITS_MAVEN_EXPECTED_NCLOCS: $(MavenExpectedNcLocs)
-              ITS_GRADLE_PIPELINE_NAME: $(AzdoItsGradlePipelineName)
-              ITS_GRADLE_SC_PROJECT_KEY: $(SonarCloudItsGradleProjectKey)
-              ITS_GRADLE_EXPECTED_COVERAGE: $(GradleExpectedCoverage)
-              ITS_GRADLE_EXPECTED_NCLOCS: $(GradleExpectedNcLocs)
-              ITS_SCANNERCLI_PIPELINE_NAME: $(AzdoItsScannerCliPipelineName)
-              ITS_SCANNERCLI_SC_PROJECT_KEY: $(SonarCloudItsCLIProjectKey)
-              ITS_SCANNERCLI_EXPECTED_COVERAGE: $(CobolExpectedCoverage)
-              ITS_SCANNERCLI_EXPECTED_NCLOCS: $(CobolExpectedNcLocs)
```
lucas-paulger-sonarsource, line 200: I am missing the rational to not use the pipeline env vars, Is it just that the values are now hardcoded in the TestCases directly?

lucas-paulger-sonarsource (approved):
LGTM

azure-pipelines.yml
7PH: Yes values are now hardcoded in test cases

Before this PR, all values were:
- Set in AzDO in a variable group, each one manually
- Forwarded to the project with env variables, each one manually (this piece of code)
- Read by the C# codebase

But:
- Each one of these variables is used in a single location, so there's no use in using variables
- We now just have to create a single file in `its/Models/` and include it in the main file to create a new E2E pipeline

lucas-paulger-sonarsource: if it's nothing secret, sounds fine to remove them. If they were somehow re-used it could also make sense but since it's not; your rational makes sense 👍🏻

PR #418, opened 2024-10-02
NO-JIRA Bump SQV6 & SCV2 patch versions to ensure users receive node-fetch revert

7PH:


lucas-paulger-sonarsource:
I was looking for context on the reason for the revert

https://community.sonarsource.com/t/prepare-analysis-configuration-fails-in-version-6-3-2/123436/17

The problem was `latest` was updated to 3.3.2 -> https://github.com/SonarSource/sonar-scanner-azdo/commit/2f5a4e9bc18081974d5fc7e0fb886fd33e6a5586

then we moved to the v7/v3 so v6 had its own versions, and had to be reverted specifically here

let me know if this is correct. Looks good to me 👍🏻

lucas-paulger-sonarsource (approved):
LGTM

7PH:
That;'s correct @lucas-paulger-sonarsource , thanks for the review, and I should have linked the ticket link to be more clear about the WHY

PR #419, opened 2024-10-07
SONARAZDO-396 Include new Scanner .NET v9 in SQ/SC extensions & Bump the embedded Scanner CLI version to v6.2.1

7PH:
- https://sonarsource.atlassian.net/browse/SONARAZDO-396
- https://sonarsource.atlassian.net/browse/SONARAZDO-410

This bumps the extension size from `24M` to `37M`, because:
- The Scanner CLI size was bumped from `4.7M` to `13M` (due to [this commit](https://github.com/SonarSource/sonar-scanner-cli/commit/8ef04a5f85ae29d14a1bfe4c930ab81e096e5456))
- The Scanner for .NET size was bumped from `2.3M`to `2.6M`

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -72,16 +72,17 @@ check_size_task:
     GPG_SIGNING_KEY: VAULT[development/kv/data/sign data.key]
     GPG_SIGNING_PASSPHRASE: VAULT[development/kv/data/sign data.passphrase]
     folder: node_modules
-    SIZE_LIMIT: 30720
+    MAX_FILE_SIZE_MB: 40
   install_script:
     - bash scripts/install.sh
   script:
     - source cirrus-env BUILD
     - npm run build
     - |
-      du -a dist/*.vsix | awk '{print $1}' | xargs -I % bash -c '
-        if [ % -ge $SIZE_LIMIT ]; then
-          echo "Error: File size exceeds limit of $SIZE_LIMIT bytes."
+      du -a -m dist/*.vsix | awk '{print $1}' | xargs -I % bash -c '
+        echo "File size: % MB"
+        if [ % -ge $MAX_FILE_SIZE_MB ]; then
+          echo "Error: File size exceeds limit of $MAX_FILE_SIZE_MB MB."
           exit 1
         fi
       '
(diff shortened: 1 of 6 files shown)


its/Models/DotnetFrameworkV3TestCase.cs
```diff
@@ -24,8 +24,8 @@ public class DotnetFrameworkV3TestCase : BuildTestCase
 	{
 		public DotnetFrameworkV3TestCase()
 		{
-			Coverage = 25.0;
-			NcLocs = 49;
+			Coverage = 2.7;
+			NcLocs = 262;
```
7PH, line 28: The .NET scanner bump changed these values

(another benefit to have them in-code rather than in variables is that we can change them here and not break master build 😛)

src/common/sonarcloud-v2/config.ts
```diff
@@ -1,6 +1,6 @@
 // When the user does not specify a specific version, these willl be the default versions used.
 const msBuildVersion = "6.2.0.85879";
-const cliVersion = "6.1.0.4477";
+const cliVersion = "6.2.1.4610";
```
7PH, line 3: Also updated minor of previous versions

.cirrus.yml
```diff
@@ -72,16 +72,17 @@ check_size_task:
     GPG_SIGNING_KEY: VAULT[development/kv/data/sign data.key]
     GPG_SIGNING_PASSPHRASE: VAULT[development/kv/data/sign data.passphrase]
     folder: node_modules
-    SIZE_LIMIT: 30720
```
lucas-paulger-sonarsource, line 75: What is the hard limit by azure devops again? I think 50 MB right? And for those on Azure DevOps Server they have a workaround for large file sizes I believe

src/common/sonarcloud-v2/config.ts
lucas-paulger-sonarsource: we can keep an eye if anyone runs into file size issues

lucas-paulger-sonarsource (approved):
LGTM 🚀

.cirrus.yml
7PH: Exactly `50M` for AzDO Services and as low as `20M` for some TFS servers (but there are work arounds)

PR #420, opened 2024-10-14
NO-JIRA Fix AzDO quality gate to ignore code coverage on non-latest SQ/SC common folders

7PH:
For `sonar(qube|cloud)-v*` folders, should:
- Stop checking coverage
- Stop checking duplication 
- Continue finding code smells/hotspots 

It means that when we create a new major task versions, if we have debt on latest, we need to accept the issues on the `sonar(qube|cloud)-v*` folders

I validated this PR with a dummy commit that I will drop after your review


lucas-paulger-sonarsource (approved):
LGTM, I see issues correctly raised for old tasks still that we will accept and duplication is not considered for old tasks

PR #423, opened 2024-10-24
SONARAZDO-413 Stop showing messages like LIB_xxx in pipeline logs

7PH:
Mitigates https://github.com/microsoft/azure-pipelines-tool-lib/issues/240 in a clean way

Merges the `lib.json` files of `azure-pipelines-task-lib` and `azure-pipelines-tool-lib` in a single `lib.json` to fix an issue [in their library](https://github.com/microsoft/azure-pipelines-tool-lib/issues/240) that doesn't properly support bundling.

This makes [all these translations](https://github.com/microsoft/azure-pipelines-task-lib/blob/7898b15726774c98a607a7e2d69898481caed201/node/lib.json#L22) available in our pipelines.

## Demo

Demo with a missing input

current extension
image

after this PR
image


lucas-paulger-sonarsource (approved):
I understand now, the “translations” from these libs were missing… I see the `lib.json` now in the task directory and correctly being used now (MUCH better now)

PR #431, opened 2024-10-29
SONARAZDO-418 Ensure systematic .NET team dogfooding before Marketpla…

lucas-paulger-sonarsource:
…ce releases

Please review our [contribution guidelines](https://github.com/SonarSource/sonar-scanner-azdo/blob/master/docs/contributing.md).

- [ ] If there is a [JIRA](https://jira.sonarsource.com/browse/SONARAZDO) ticket available, please make your commits and pull request start with the ticket ID (SONARAZDO-XXXX)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,6 +16,8 @@ variables:
     value: true
   - name: NPM_CONFIG_registry
     value: $(ARTIFACTORY_URL)/api/npm/npm
+  - name: isMain
+    value: $[eq(variables['Build.SourceBranch'], 'refs/heads/master')]
 
 stages:
   - stage: build
@@ -62,7 +64,7 @@ stages:
               customCommand: "run test-build -- --publisher $(publisher)"
 
           - task: CopyFiles@2
-            displayName: "Copy artefacts to build directory"
+            displayName: "Copy artifacts to build directory"
             inputs:
               SourceFolder: "$(Build.SourcesDirectory)/dist"
               Contents: "*.vsix"
@@ -89,10 +91,13 @@ stages:
       - job: "run_qa"
         variables:
           publisher: $(SonarSourceITPublisher)
-          sqExtensionVersion: $[ stageDependencies.build.npminstall.outputs['versions.SQ_VERSION'] ]
-          scExtensionVersion: $[ stageDependencies.build.npminstall.outputs['versions.SC_VERSION'] ]
+          sqExtensionVersion: $[stageDependencies.build.npminstall.outputs['versions.SQ_VERSION']]
+          scExtensionVersion: $[stageDependencies.build.npminstall.outputs['versions.SC_VERSION']]
           itOrg: $(SonarSourceITOrganization)
         steps:
+          - bash: |
+              echo "##vso[task.setvariable variable=SC_VERSION;isoutput=true]$(scExtensionVersion)"
+            name: versions
           # Unpublish extension
           - task: TfxInstaller@3
             displayName: "Install TFX CLI"
@@ -131,6 +136,7 @@ stages:
               downloadType: "single"
               artifactName: "extensions"
               downloadPath: "$(System.ArtifactsDirectory)"
+
           - task: PublishAzureDevOpsExtension@3
             displayName: "Publish SonarCloud test extension"
             inputs:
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -16,6 +16,8 @@ variables:
     value: true
   - name: NPM_CONFIG_registry
     value: $(ARTIFACTORY_URL)/api/npm/npm
+  - name: isMain
+    value: $[eq(variables['Build.SourceBranch'], 'refs/heads/feature/lp/create-dogfooding-extension')] # replace with refs/heads/master
```
lucas-paulger-sonarsource, line 20: after validating this check works, I will change to `refs/head/master`

azure-pipelines.yml
```diff
@@ -185,3 +189,45 @@ stages:
             inputs:
               command: "test"
               projects: "its/runner/IntegrationTests.csproj"
+
+  - stage: publish_dogfood
+    displayName: "Publish dogfood extension"
+    condition: and(succeeded(), eq(variables['isMain'], true))
```
7PH, line 195: NP / disclaimer: not tested

```suggestion
    condition: and(succeeded(), variables['isMain'])
```

azure-pipelines.yml
```diff
@@ -185,3 +189,45 @@ stages:
...
+    jobs:
+      - job: "publish_dogfood"
+        variables:
+          publisher: $(SonarSourceITPublisher)
+          scExtensionVersion: $[stageDependencies.build.npminstall.outputs['versions.SC_VERSION']]
+          deps: $[convertToJson(dependencies)] # to debug if dependencies are correct
```
7PH, line 201: I don't understand where this comes from?

azure-pipelines.yml
```diff
@@ -185,3 +189,45 @@ stages:
...
               projects: "its/runner/IntegrationTests.csproj"
+
+  - stage: publish_dogfood
+    displayName: "Publish dogfood extension"
+    condition: and(succeeded(), eq(variables['isMain'], true))
+    jobs:
+      - job: "publish_dogfood"
+        variables:
+          publisher: $(SonarSourceITPublisher)
+          scExtensionVersion: $[stageDependencies.build.npminstall.outputs['versions.SC_VERSION']]
```
7PH, line 200: I think it fails because this variable comes from another stage, wdyt?

gulpfile.js
```diff
@@ -497,7 +497,7 @@ gulp.task("ci:azure:get-extensions-version", (done) => {
     console.log(`Fetched ext version ${version} for ${extension}`);
     console.log(`##vso[task.setvariable variable=${extensionPrefix}_EXT_NAME]${extensionManifest}`);
     console.log(
-      `##vso[task.setvariable variable=${extensionPrefix}_VERSION;isOutput=true]${version}`,
+      `##vso[task.setvariable variable=${extensionPrefix}_VERSION;isOutput=true;isreadonly=true]${version}`,
```
lucas-paulger-sonarsource, line 500: I was thinking it was getting overridden between stages, but as I found out, only the last stage is passed to the `stageDependencies`. I can omit this but in theory it is read only 🤷🏻 
https://learn.microsoft.com/en-us/azure/devops/pipelines/process/set-variables-scripts?view=azure-devops&tabs=bash#set-variable-properties

azure-pipelines.yml
```diff
@@ -185,3 +187,40 @@ stages:
...
+              connectTo: "VsTeam"
+              connectedServiceName: "AzDO Extension - Marketplace - Dogfood"
+              fileType: "vsix"
+              vsixFile: "$(System.ArtifactsDirectory)/extensions/sonar-scanner-azdo-$(scExtensionVersion)-sonarcloud.vsix"
+              extensionName: "[Dogfood] SonarCloud"
+              extensionId: "sonarcloud-dogfood"
+              updateTasksVersion: false
+              updateTasksId: false
+              extensionVisibility: "private"
+              shareWith: "sonarsource"
```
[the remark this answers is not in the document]
7PH: I agree, like it inline now TBH

7PH (approved):
We don't have the permission to install the extension on the `SonarSource` organization, so that'll be another thing to check

LGTM ✅
image

azure-pipelines.yml
lucas-paulger-sonarsource: I'll keep it as is in that it clearly states what we expect (boolean true)
examples shown here
https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#stage-to-stage-dependencies

azure-pipelines.yml
lucas-paulger-sonarsource: addressed and works

PR #432, opened 2024-10-29
SONARAZDO-416 Auto-generate test pipelines & Add dummy .NET Core, Framework, CLI, Gradle and Maven projects with expected nloc/issues

7PH:
[ticket](https://sonarsource.atlassian.net/browse/SONARAZDO-416)


its/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  for (const combination of generateCombinations()) {
+    const filename = `${FILENAME_PREFIX}-${serializeCombination(combination)}`;
+    const pipeline = generatePipelineFile(combination);
+
+    const pipelinePath = path.join(FIXTURES_PATH, `${filename}.yml`);
+    console.log(`Writing ${pipelinePath}`);
+    fs.writeFileSync(pipelinePath, pipeline);
+  }
+}
+
```
lucas-paulger-sonarsource, line 28: I think if they are generated as part of the pipeline, we should ignore them from git

its/pipeline.ts
```diff
@@ -0,0 +1,188 @@
...
+import { PipelineCombination, TaskDefinition, YamlContent } from "./types";
+
+function generatePipelineHeader(config: PipelineCombination): YamlContent {
+  return {
+    trigger: "none",
+    pool: {
+      vmImage: config.os === "unix" ? "ubuntu-latest" : "windows-latest",
+    },
+    variables: {
+      "system.debug": true,
```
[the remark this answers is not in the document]
7PH: It is useful IMO to debug when they fail in CI, I don't think there's a performance hit because of this, but I'll check

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "BUILD_MODE=production gulp",
     "test-build": "gulp",
+    "build-its": "ts-node its/index.ts",
```
lucas-paulger-sonarsource, line 9: This requires manually running these to update the IT configurations. 

It could be beneficial to run this as a CI job too to catch any changes

its/fixtures/pipeline-sonarcloud-v2-other-maven-unix.yml
lucas-paulger-sonarsource, line 1: why don't I see this locally? maven/gradle are excluded from the name 🤔 
Screenshot 2024-10-30 at 12 24 19

its/fixtures/dummy-project-cli/.gitignore
lucas-paulger-sonarsource, line 1: are these fixture project code coming from anywhere in particular? like the repos on azdo 
https://dev.azure.com/sonarsource-integration-tests/_git/sonarsource-test-project

its/index.ts
7PH: As agreed, I'll drop them and try to find an automated solution right away

its/fixtures/dummy-project-cli/.gitignore
7PH: I've created new dummy clean repos with recent versions of gradle/maven/dotnet core/framework with each with only a few nloc/issues so they don't take long building/analyzing

its/fixtures/pipeline-sonarcloud-v2-other-maven-unix.yml
7PH: weird since [they're here](https://github.com/SonarSource/sonar-scanner-azdo/blob/9e1a2245114e60102a5fb85cfb755e9c986d4bda/its/fixtures/pipeline-sonarcloud-v1-other-gradle-unix.yml) 🤔

package.json
7PH: I've created a [subtask](https://sonarsource.atlassian.net/browse/SONARAZDO-428) for this although I'm not certain for now how to implement it

its/index.ts
7PH: (deleted my previous comment)
As discussed on Slack,  in order to be able to run them in AzDO, they need to be generated and available in the master branch

PR #438, opened 2024-11-06
NO-JIRA Fix master QA pipeline

7PH:
Fixes the master QA pipeline

### DIFF
--- a/its/it/cases.ts
+++ b/its/it/cases.ts
@@ -10,7 +10,6 @@ type TestCase = {
   sonarHostUrl: string;
   projectKey: string;
   pipelineName: string;
-  nloc: number;
 };
 
 export const testCases: TestCase[] = [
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -27,6 +27,8 @@ variables:
     value: $(ARTIFACTORY_URL)/api/npm/npm
   - name: isMain
     value: $[eq(variables['Build.SourceBranch'], 'refs/heads/master')]
+  - name: azureBranch
+    value: $[coalesce(variables['System.PullRequest.SourceBranch'], 'master')]
```
7PH, line 31: Turns out we must use a pre-defined variable, otherwise it doesn't work if we directly put it in the `env` section of a task

lucas-paulger-sonarsource (approved):
🚀

PR #440, opened 2024-11-07
SONARAZDO-416 Use snapshot-based testing to ensure generated pipelines did not unexpectedly changed without being committed

7PH:
I think a simple solution like this is enough to ensure the generate pipelines are always up to date 

If there was an uncommitted change to the generates pipelines, the Cirrus build fails and shows the diff
image

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -54,6 +54,7 @@ build_task:
   script:
     - git fetch
     - source cirrus-env BUILD
+    - npm run generate-its && git diff --exit-code # Ensure that the generated ITS are up-to-date
     - npm run validate-ci
     - npm run build
     - npm run sonarqube


.cirrus.yml
```diff
@@ -54,6 +54,7 @@ build_task:
   script:
     - git fetch
     - source cirrus-env BUILD
+    - npm run generate-its && git diff --exit-code # Ensure that the generated ITS are up-to-date
```
lucas-paulger-sonarsource, line 57: very cool! I like it. One question, when the changes are intentional; the changes will be committed so **this** `git diff --exit-code` should still pass, only when the changes aren't committed 👍🏻

lucas-paulger-sonarsource (approved):

.cirrus.yml
7PH: Yes exactly 👍

PR #442, opened 2024-11-07
SONARAZDO-416 Run the full QA on master branch only

7PH:
I'm adding you to review but keeping the PR as draft, because I'm not sure we want to merge this @lucas-paulger-sonarsource


lucas-paulger-sonarsource:
This is fine when its just SC, but with SQ and SQ LTS, we add a lot more cases. But less configuration overhead for the test cases. We can keep the PR and think on it IMO

7PH:
How about I modify the code so only the nightly runs the full QA, and the master + PR tests still runs what they run today? Do you think that'd make sense?

PR #448, opened 2024-11-14
NO-JIRA Bump latest extension/task versions to 3.1.0 (SQC) and 7.1.0 (SQS)

7PH:
- Bump latest SonarQube Server extension versions to `7.1.0`
- Bump latest SonarQube Cloud extension versions to `3.1.0`

So that users get the rebranded extension/tasks


lucas-paulger-sonarsource (approved):
> So that users get the rebranded extension/tasks

I didn't know a patch isn't always updated, is that from experience or from docs?

7PH:
As discussed, posting for clarity - both `patch` and `minor` are automatically upgrading user pipelines

PR #453, opened 2024-11-21
SONARAZDO-440 Forward the cliSources input as intended to the CLI Scanner

7PH:
[ticket](https://sonarsource.atlassian.net/browse/SONARAZDO-440)

## Changes

What this PR changes is that the RAW value of `cliSources` is now used instead of the converted value to an absolute path:
image (before)
vs
image (now)

## Demo

Assuming a repository structure like this:
```text
- docs-src/src/components
```

With this setup:
```yml
- task: SonarCloudPrepare@3
  inputs:
    [..]
    cliSources: src\components
    extraProperties: |
      sonar.projectBaseDir=$(Build.SourcesDirectory)\docs-src
```

The value of `cliSources` is changed before it is setn to the task to be an absolute path starting from the repository root (not `projectBaseDir`). Therefore the analysis fails

image

With this fix, with the same task definition, we can see the results on SonarQube Cloud which is as expected

image

### DIFF
--- a/src/extensions/sonarcloud/tasks/SonarCloudPrepare/v3/task.json
+++ b/src/extensions/sonarcloud/tasks/SonarCloudPrepare/v3/task.json
@@ -145,7 +145,7 @@
     },
     {
       "name": "cliSources",
-      "type": "filePath",
+      "type": "string",
       "label": "Sources directory root",
       "defaultValue": ".",
       "required": true,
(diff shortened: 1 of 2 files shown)


lucas-paulger-sonarsource (approved):
thanks for the detailed explanation 🙇🏻

PR #465, opened 2025-04-03
SONARAZDO-462 Optimize QA pipeline

henryju:
[SONARAZDO-462](https://sonarsource.atlassian.net/browse/SONARAZDO-462)

Before:
image

After:
image

[SONARAZDO-462]: https://sonarsource.atlassian.net/browse/SONARAZDO-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/combination.ts
+++ b/its/combination.ts
@@ -1,12 +1,25 @@
 import { PipelineCombination } from "./types";
 
 export function generateCombinations(): PipelineCombination[] {
-  const osCombinations: PipelineCombination["os"][] = ["unix", "windows"];
-  const versionCombinations: PipelineCombination["version"][] = [
-    { extension: "sonarcloud", version: 1 },
-    { extension: "sonarcloud", version: 2 },
-    { extension: "sonarcloud", version: 3 },
-  ];
+  let osCombinations: PipelineCombination["os"][];
+  switch(process.env.RUN_ON_OS) {
+      case "unix": osCombinations = ["unix"]; break;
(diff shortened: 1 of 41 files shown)


its/combination.ts
```diff
@@ -1,12 +1,25 @@
...
 
 export function generateCombinations(): PipelineCombination[] {
-  const osCombinations: PipelineCombination["os"][] = ["unix", "windows"];
-  const versionCombinations: PipelineCombination["version"][] = [
-    { extension: "sonarcloud", version: 1 },
-    { extension: "sonarcloud", version: 2 },
-    { extension: "sonarcloud", version: 3 },
-  ];
+  let osCombinations: PipelineCombination["os"][];
+  switch(process.env.RUN_ON_OS) {
```
antoine-vinot-sonarsource, line 5: I don't understand where those env variables are coming from or where they are defined. How did you find it?

antoine-vinot-sonarsource (approved):
LGTM! Nice to have also added the cache tasks!
I just have a question for my knowledge about the env variables used in combination

its/combination.ts
henryju: I used to pass them from the Azure Pipeline when I tried the matrix job. Then I reverted to a single job (see details in the JIRA ticket) but I decided to keep this code in case we want to run only a subset of the tests.

PR #484, opened 2025-07-30
SONARAZDO-452 Add Node20 Executor

fiveshotsofespresso:


cmandesign:
I believe this is a good PR, will fix warnings on node version and also it will use the LTS version instead of non-LTS one.

colin-sonarsource:
Based on official Azure DevOps Tasks, we're safe to have an agent version that might not necessarily support the new executor in order to preserve backwards compatability.

https://github.com/microsoft/azure-pipelines-tasks/blob/3e143d0b011e6c130c230aab627b28727a6c53b5/Tasks/NotationV0/task.json#L231

We have also had users forcing Node20 execution (via env variable) for a while with no reports of issues. Unfortunately that env variable doesn't seem to supress the task warnings polluting pipelines.

We should test the extension manually to be extra sure it gets rid of the warnings, and doesn't have any unintended side effects.

antoine-vinot-sonarsource (approved):
Thank you Colin for the changes!
I tested it and the warnings are gone 🚀 
azure_no_node16_warnings

antoine-vinot-sonarsource:
@MarkvanOsch 
Hi, we plan to release today.

PR #583, opened 2026-06-05
SC-49820 Severity metrics decoration

OrlovAlexander85:
Adding formatters for new Severity Metrics Conditions


src/common/latest/helpers/measures.ts
```diff
@@ -142,6 +172,26 @@ function comparatorFormatter(value: string): string {
   return l10nKeys[value.toUpperCase()] ?? value;
 }
 
+function severityFormatter(value: string | number): string {
+  const numVal = typeof value === "string" ? Number.parseInt(value, 10) : value;
+  return STANDARD_SEVERITY_LABELS[numVal] ?? value.toString();
```
stanislavhh: Can typecast to number,  this will also return 0 when value is an empty string `''` and seems like this is more desired:
```suggestion
  return STANDARD_SEVERITY_LABELS[Number(value)] ?? value.toString();
```

src/common/latest/sonarqube/__tests__/HtmlAnalysisReport-test.ts
```diff
@@ -122,3 +122,111 @@ it("should display the project name", () => {
...
+            comparator: "GT",
+            errorThreshold: threshold,
+            actualValue,
+          },
+        ],
+      },
+      [],
+      {
+        ...MOCKED_ANALYSIS_RESULT,
+        metrics: [{ key: metricKey, name: "Severity", type: "INT" }],
```
stanislavhh, line 168: could we provide metrics for severity here?  This would help to better understand/validate how the html looks like. Specifically, would be nice to understand what is  the whole string for measure and its comparator

stanislavhh (commented):
Overall looks good, small suggestion,  and given that we introduced new measure type, it will be a good idea to verify that measure is rendered in html as expected (comparator, measure name)

stanislavhh (approved):
LGTM

src/common/sonarcloud-v3/helpers/measures.ts
```diff
@@ -20,6 +20,34 @@
 
 const HOURS_IN_DAY = 8;
 
+const STANDARD_SEVERITY_LABELS: { [value: number]: string } = {
```
antoine-vinot-sonarsource: Writing here, but this is a global comment for `sqc-v3` and `sqs-v7` folders:
v3 and v7 are deprecated maintained versions of the extension tasks.
Do we really want to bring this new update to the previous task versions? Or should we focus on latest only?

src/common/latest/helpers/measures.ts
```diff
@@ -20,6 +20,34 @@
...
+  "new_bugs_severity",
+  "new_vulnerabilities_severity",
+  "new_code_smells_severity",
+]);
+
+const MQR_SEVERITY_METRIC_KEYS = new Set([
+  "new_software_quality_reliability_severity",
+  "new_software_quality_security_severity",
+  "new_software_quality_maintainability_severity",
+]);
```
antoine-vinot-sonarsource, line 49: I'm a bit concerned that we tightly couple the AzDO extensions with these metrics key and mapping. It looks like a fragile design.
It can break what we show if we, for example, decide to change these mappings on server side and forget to update it here. 
For example, say we remove one severity level on server side and forget to change it here as well.
Same if we change, or add, metric keys in the future, this will also require an update here, which can be easily forgotten.

src/common/sonarcloud-v3/helpers/measures.ts
OrlovAlexander85: I am not sure how different versions are used and what is the support strategy, but if we do not include this mapping the decoration will show confusing numeric values. Iw that is expected behavior for v3 and v7, i can remove those changes.

antoine-vinot-sonarsource: I suggest we remove these changes from the v3 and v7 tasks. If people are still on these versions, yes they will see confusing values. However, they should still upgrade to v4 or v8, our doc explicitly states v3 and v7 are deprecated.

src/common/latest/helpers/measures.ts
OrlovAlexander85: Valid concern, but on the other hand, we have 5 level severity for quite some time, so I doubt it will change anytime soon, and If it will we will need to update this class.

antoine-vinot-sonarsource: IMO the AZDO extension should not have to worry about that and just use what it's receiving from the metric/measure endpoints. That's really a shame that we must have this coupling.
That being said, I understand that we don't have many other options at the moment. I mean, I can't think of any straightforward one.
So I'm OK to proceed with this change for now.

antoine-vinot-sonarsource (commented):
@OrlovAlexander85 Quick heads-up. I'm trying to validate the changes from a functional point of view, because our ITs don't handle the fail QG scenario I believe. However I faced some issues (not related to your changes) when trying it on a Dev environment. 
Please bear with me a little bit more 🙏

antoine-vinot-sonarsource (changes requested):
I tried it out and the change works for the formatted evaluation at the start of the line. However, the required threshold is not properly formatted because it shows "one less" than the value associated to the given severity.
See my screenshot below: I set up the QG to INFO level, hence it show `required <= 4` and not `required < INFO` (Or whatever we want there)

Image

OrlovAlexander85:
@antoine-vinot-sonarsource Thanks a lot for doing functional validation. 
I will add a proper mapping to a condition as well

antoine-vinot-sonarsource (approved):
LGTM!

Image

@OrlovAlexander85 The pipeline failed because of some timeout when publishing the test extension. It's a known thing for which I did not find a solution yet sadly. You might need to retrigger a new pipeline, as rerunning the failed one won't work I think.

OrlovAlexander85:
Hey @antoine-vinot-sonarsource , once again thanks a lot for validation. I rerun the pipeline and all checks are green and PR is ready to be merged.
Can you please merge it, as I do not have required permissions on this repo.

PR #585, opened 2026-06-18
SONARAZDO-589 Fix test using promise-returning assertion without awaiting

antoine-vinot-sonarsource:
Fixes issues appearing after new rules were added to the Typescript QP.

----
## Summary by Gitar

- **Test fixes:**
  - Added `async/await` to tests in `api-test.ts` to properly handle promise-returning assertions
  - Corrected `expect(...).rejects.toThrow()` calls that were previously missing the required `await`

<sub>This will update automatically on new commits.</sub>

### DIFF
--- a/src/common/latest/helpers/__tests__/api-test.ts
+++ b/src/common/latest/helpers/__tests__/api-test.ts
@@ -122,18 +122,18 @@ describe("fetchProjectStatus", () => {
     expect(projectStatus.conditions).toEqual(MOCKED_CONDITIONS);
   });
 
-  it("should fail if the project status fails to load", () => {
+  it("should fail if the project status fails to load", async () => {
     jest.mocked(get).mockRejectedValueOnce(new Error("API Couldn't be reached"));
 
-    expect(() => fetchProjectStatus(MOCKED_ENDPOINT, "analysisId")).rejects.toThrow(
+    await expect(fetchProjectStatus(MOCKED_ENDPOINT, "analysisId")).rejects.toThrow(
       "Could not fetch analysis for ID 'analysisId'",
     );
   });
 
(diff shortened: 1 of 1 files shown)


zglicz (approved):
The change makes sense: for Jest `.rejects`, the assertion promise should be awaited, while the promise under test should be passed directly to `expect(...)`.

One follow-up worth considering: the same unawaited `.rejects` pattern is still present in the sibling common variants:
- `src/common/sonarqube-v7/helpers/__tests__/api-test.ts:125`, `:133`, `:185`, `:198`, `:205`
- `src/common/sonarcloud-v3/helpers/__tests__/api-test.ts:125`, `:133`, `:185`, `:198`, `:205`

Those look like active test packages under the root `npm test` wiring, so it would be good to apply the same cleanup there too, either in this PR or a follow-up.

antoine-vinot-sonarsource:
> The change makes sense: for Jest `.rejects`, the assertion promise should be awaited, while the promise under test should be passed directly to `expect(...)`.
> 
> One follow-up worth considering: the same unawaited `.rejects` pattern is still present in the sibling common variants:
> 
> * `src/common/sonarqube-v7/helpers/__tests__/api-test.ts:125`, `:133`, `:185`, `:198`, `:205`
> * `src/common/sonarcloud-v3/helpers/__tests__/api-test.ts:125`, `:133`, `:185`, `:198`, `:205`
> 
> Those look like active test packages under the root `npm test` wiring, so it would be good to apply the same cleanup there too, either in this PR or a follow-up.

@zglicz Thank you for the review! 
v7 and v3 are deprecated versions of our tasks. I made a deliberate choice to not touch them and accepted the related issues.

PR #586, opened 2026-06-18
SONARAZDO-590 Prepare sq-8.2.3 and sc-4.2.5 release

antoine-vinot-sonarsource:


zglicz (approved):
Why does it need to be spread out through so many files? Aren't some of these values derived from others and could be automated to keep them in sync?

antoine-vinot-sonarsource:
> Why does it need to be spread out through so many files? Aren't some of these values derived from others and could be automated to keep them in sync?

@zglicz 

**Short answer**: It's how Azure DevOps works. 

**Long answer**: The extension version is defined in the `vss-extension.json` file (we maintain two extensions in this project, sqs and sqc). Each extension can have multiple tasks with their own version. A task is an independently versioned unit in AZDO. So in theory it's possible to have different numbers, even though in practice the values aren't truly independent per convention. 
I created an [update-version.sh utility script](https://github.com/SonarSource/sonar-scanner-azdo/blob/893d07e810d25db146f4a1797304eed3577981f3/scripts/update-version.sh) to not have to manually edit this every time. But I agree there is probably plenty of room for improvement with this process

PR #598, opened 2026-08-04
SONARAZDO-603 Workaround for request timeout issues on extension publish step

antoine-vinot-sonarsource:
## Validation

The pipeline ran with only a few seconds to publish the test extension. Most time is spent validating the extension, in our example there was 5 attempts. (cf. screenshots below)

azdo1
azdo2

----
## Summary by Gitar

- **CI/CD Pipelines:**
    - Added `noWaitValidation: true` to extension publish tasks to avoid transient timeouts
    - Introduced dedicated `IsAzureDevOpsExtensionValid@5` validation steps with retry logic

<sub>This will update automatically on new commits.</sub>

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -166,7 +166,25 @@ stages:
               updateTasksId: false
               extensionVisibility: "private"
               shareWith: "$(itOrg)"
+              # Let tfx exit right after the upload. Waiting for Marketplace validation inside the
+              # publish command makes a transient timeout fail a step that cannot safely be retried.
+              noWaitValidation: true
               cwd: "$(System.ArtifactsDirectory)"
+
+          # Wait for validation in a dedicated, read-only step that is safe to retry
+          - task: IsAzureDevOpsExtensionValid@5
+            displayName: "Wait for Marketplace validation of the test extension"
+            retryCountOnTaskFailure: 3
+            inputs:
+              connectTo: "AzureRM"
+              connectedServiceNameAzureRM: "$(MarketplaceARMServiceConnection)"
+              method: "id"
+              publisherId: "$(publisher)"
+              extensionId: "sonarcloud-test"
+              extensionVersion: "$(scExtensionVersion)"
+              maxRetries: "10"
+              minTimeout: "0.5"
+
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -166,7 +166,25 @@ stages:
               updateTasksId: false
               extensionVisibility: "private"
               shareWith: "$(itOrg)"
+              # Let tfx exit right after the upload. Waiting for Marketplace validation inside the
+              # publish command makes a transient timeout fail a step that cannot safely be retried.
+              noWaitValidation: true
               cwd: "$(System.ArtifactsDirectory)"
+
+          # Wait for validation in a dedicated, read-only step that is safe to retry
+          - task: IsAzureDevOpsExtensionValid@5
```
dorian-burihabwa-sonarsource, line 175: I am not too familiar with AZDO conventions: Is `IsAzureDevOpsExtensionValid` define somewhere in our code? Or does it live a marketplace?

dorian-burihabwa-sonarsource (approved):
LGTM but because I only so much about Azure Devops, I had a few questions:
1. Where is the task defined?
2. How can we test that is working properly?

azure-pipelines.yml
antoine-vinot-sonarsource: It's an [Azure Devops extension task](https://github.com/microsoft/azure-devops-extension-tasks/blob/9d3a1016a1e6eef3f95ba65b937d151e1e2b24e8/BuildTasks/IsValidExtensionAgent/task.json#L3).

antoine-vinot-sonarsource:
> LGTM but because I only so much about Azure Devops, I had a few questions:
> 
> 1. Where is the task defined?
> 2. How can we test that is working properly?

@dorian-burihabwa-sonarsource 
Yes sorry, I throw you under the bus here ^^'
This is purely a change in the CI pipeline (it's a bit "similar" to GitHub workflows), so to validate we just need to have a look in the pipeline and verify what happens there. I attached screenshots of the last run for this PR at the top of the PR, I don't know if you can connect to our Azure Devops repository yet.

